PR #71, opened 2014-10-30
add upstart init

wilreichert:
Native upstart service for Ubuntu, it behaves identically to the sysvinit script.  Anyone upgrading will get blessed with 2 init files which might cause a bit of confusion.


lyrixx:
Hi, Sorry to "spam" this thread, but what id the right procedure to migrate ?

PR #91, opened 2014-11-18
Enable statsd

jdef:


adgaudio:
+1 would be really great to merge #114 and #115

PR #92, opened 2014-11-20
The service_user and group option were being locally overriden in the service

ericfode:
recipe and causing problems when being used with watch resources on centos. 
the workaround was to set the user and group explicitly to root.

To fix it i moved the logic for determining the user to use out to the 
attributes file.

### DIFF
--- a/recipes/_service.rb
+++ b/recipes/_service.rb
@@ -29,6 +29,9 @@ if node['consul']['init_style'] == 'runit'
   consul_directories << '/var/log/consul'
 end
 
+consul_user  = node['consul']['service_user']
+consul_group = node['consul']['service_group']
+
 # Create service user
 user "consul service user: #{consul_user}" do
   not_if { consul_user == 'root' }


donaldguy:
@johnbellone This does not work, for wrapper cookbooks at least. init_style isn't set to runit until after compile of the cookbook, by which the user and group are already hardcoded as root on both node and in recipe variables

I am (working on) working around by using a chef-sugar compile_time block, but the better option might be to do the conditional evaluation in a ruby_block resource at runtime and user lazy attributes, OR just go all the way to LWRP territory

donaldguy:
yeah, no luck doing
```
compile_time do
  node.default['consul']['init_style'] = 'runit'
end
```

I think I'm just gonna have to manually (re)set `service_user` and `service_group` myself, but even that may probably has to be done at compile time to propagate from a wrapper cookbook and I'm not 100% that will work either (will report back after testing)

We largely follow the patterns layed out by @reset in his chefconf talk last year (https://www.chef.io/blog/chefconf-talks/the-berkshelf-way-jamie-winsor/ ) ... so I guess I might just have to deviate from that slightly and stick it as an environment override... ? (neither Berkshelf nor Policyfiles seems to provide a place to persist that at the right place in the load order... we can start passing dna.json's around but I'd rather not)

PR #103, opened 2015-01-05
Add Service Watch LWRP

monkeylittle:
Small addition to support the definition of a service watch.  The same conventions used for event and key watch have been applied.


ericfode:
:+1:

PR #104, opened 2015-01-06
fix for start then restart

zeroXten:
Had a problem converging cluster nodes once a bootstrap node existed. Essentially it was doing this:

    service consul start & sleep 3 && service consul restart
    [1] 6031
    Starting consul
    [1]+  Done                    service consul start
    Stopping consul.............
    consul not stopped; may still be shutting down or shutdown may have failed
    Unable to stop consul, will not attempt to start

action [:start, :enable] was starting the consul process, which was followed immediately by a :restart which would fail.

[Review on Reviewable](https://reviewable.io/reviews/johnbellone/consul-cookbook/104)

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -4,7 +4,7 @@ maintainer_email 'jbellone@bloomberg.net'
 license 'Apache v2.0'
 description 'Installs/Configures Consul client, server and UI.'
 long_description 'Installs/Configures Consul client, server and UI.'
-version '0.6.0'
+version '0.6.1'
 
 recipe 'consul', 'Installs and starts consul service.'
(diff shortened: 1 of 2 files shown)


zeroXten:
eh, hang on. This might not be working as desired.

metadata.rb
```diff
@@ -4,7 +4,7 @@
 license 'Apache v2.0'
 description 'Installs/Configures Consul client, server and UI.'
 long_description 'Installs/Configures Consul client, server and UI.'
-version '0.6.0'
+version '0.6.1'
```
johnbellone: Please don't bump version in a pull-request.

PR #108, opened 2015-01-13
Make upstart script respawn consul on crash

tgwizard:
This could be configurable (whether to respawn, and the count/interval/unlimited), but I think it could be a future PR.

Is there a reason to not respawn if consul crashes?

### DIFF
--- a/templates/default/consul.conf.erb
+++ b/templates/default/consul.conf.erb
@@ -18,3 +18,6 @@ end script
 post-start exec initctl emit consul-up
 
 kill signal INT
+
+respawn
+respawn limit 10 10


lyrixx:
:+1:  for the features

tgwizard:
@johnbellone what do you think?

johnbellone:
I personally don't see anything wrong with this. Since its a change in behavior we should just bump the minor revision.

Any thoughts @reset ?

johnbellone:
@tgwizard Is there any logic in Consul that is destructive on a restart? Assuming it crashed we want to make sure that we're not wiping out data. Otherwise, it :+1: !

tgwizard:
@johnbellone there seems to be no issue for me. I can `sudo kill -9 $PID`, the consul process dies, and then restarts and rejoins the cluster. I have `rejoin_after_leave` set to `true` in my consul config though, which is (I think) what causes the rejoin to work. Or perhaps not, since no leave was triggered.

PR #117, opened 2015-02-02
fix #116 by replacing :touch with :create action

### LINKED ISSUE #116: install_binary breaks upgrade
The file resource in install_binary which is intended to change the permissions of the installed binary also changes the mod_date of the executable to the current date. E.g.: if I install an older version of consul in 2014 December, which has mod_date of 2014 August, the file resource will change the mod_date to 2014 December. Now if I want to install a newer version that was released e.g in 2014 November, the unzip step of the ark resource will not overwrite the older version executable as it has newer mod_date than the newer version (2014 Dec vs 2014 Nov).

akerekes:
These changes were introduced by #91 (among others). I've split that PR into smaller chunks in hope for easier merge.

### DIFF
--- a/recipes/install_binary.rb
+++ b/recipes/install_binary.rb
@@ -32,6 +32,5 @@ end
 consul_file = File.join(node['consul']['install_dir'], 'consul')
 file "#{consul_file}" do
   mode '0755'
-  action :create
   only_if { File.exists?("#{consul_file}") }
 end
--- a/recipes/install_binary.rb
+++ b/recipes/install_binary.rb
@@ -29,7 +29,9 @@ ark 'consul' do
   action :dump
 end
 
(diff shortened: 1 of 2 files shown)


tgwizard:
Really :+1: for this! The failing tests seem to be a common problem of all PRs of late, `ChefSpec::Error::CommandNotStubbed:`, not a problem with this specific change.

johnbellone:
I can take a look into the failed tests separately. 

Can we remove the action statement since the file resource uses create by default? It's pretty confusing looking at it with the guard in there.

johnbellone:
Please rebase!

akerekes:
#116 has been fixed by #121, so I think this pull request can be closed.

PR #118, opened 2015-02-02
Set GOMAXPROCS > 1

tgwizard:
This it should be, according to https://groups.google.com/forum/#!topic/consul-tool/qewFEqgAoF8. This PR still makes it possible to set it to one, just changes the default to `max(node['cpu']['total'], 2)`.

One issue with this change though is that the consul service isn't restarted when the change has been applied to `/etc/default/consul`. A `:restart` notification could be added for that resource, but then the service would also be restarted once on the initial install. I'm not sure how this should be solved, and if.


.ruby-gemset
```diff
@@ -0,0 +1 @@
+consul-cookbook
```
johnbellone: Please remove the ruby-gemset and ruby-version files in root of the repository.

attributes/default.rb
```diff
@@ -43,6 +43,14 @@
 }
 default['consul']['source_revision'] = 'master'
 
+# GOMAXPROCS should be greater than 1, see
+# https://groups.google.com/forum/#!topic/consul-tool/qewFEqgAoF8
+if node['cpu']['total'] > 1
+  default['consul']['gomaxprocs'] = node['cpu']['total']
```
johnbellone: So the problem that we're trying to solve here is if there is only a single CPU in an instance that we set this greater than one?

johnbellone: I guess I am just not sold that this needs to be able to be tuned as a node attribute. Perhaps we can set it as a variable in the template instead of rendering the attribute? /cc @reset any thoughts?

tgwizard: Well, the only thing I wanted to change is that it isn't set to 1 by default, which (according to the mailing list thread in the comment) isn't completely safe.

.ruby-gemset
tgwizard: Certanly, sorry! Would it be fine if I add them to .gitignore instead, so that I can use them locally?

reset: @tgwizard yeah, it's just fine to add them to gitignore

tgwizard: Great, done!

attributes/default.rb
johnbellone: If that's the case I would prefer this to be in the template.

```ruby
GOMAXPROCS=<%= [ node['cpu']['total'], 2 ] %>
```

johnbellone:
Please rebase!

PR #121, opened 2015-02-05
Upgradeable Consul Binary

reset:
This PR adds the ability to upgrade or downgrade the installed Consul binary to a different version than already present on the node.

Partial convergeance errors generated by the cookbook will also be remedied as a side effect from changing our archive management dependency from 'ark' to the more reliable and and 100% idempotent cousin - libarchive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,12 +323,6 @@ Following attributes, if exist in the [encrypted databag][7], override the node 
     <td><tt>nil</tt></td>
   </tr>
   <tr>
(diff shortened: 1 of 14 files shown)


johnbellone:
I like what I am seeing!

reset:
@johnbellone this is complete now. It's best to give the cookbook a new minor version or (not major since we're 0.x still). I think we should 1.0 once we've cleaned up the code a bit. When we chat tomorrow I'll tell you what I'm thinking ;).

PR #123, opened 2015-02-06
correction to EL init template for active binary and config dir argument...

paulysullivan:
...s to agent


reset:
@paulysullivan thanks, I pushed this change out in it's own patch release (0.8.1) since it was such a major oversight

PR #126, opened 2015-02-08
Add HWRPs for installing and managing consul.

johnbellone:
I am channeling @someara and @reset here.

After taking a look at the new [httpd library cookbook][1] as well
as the [Mysql library cookbook][2] I think this is an excellent example
of using HWRP for defining the building blocks of a quality cookbook.

[1]: https://github.com/chef-cookbooks/httpd
[2]: https://github.com/chef-cookbooks/mysql

PR #127, opened 2015-02-09
Delay return of control during startup

### LINKED ISSUE #125: Sensitivity to HUP during launch
The consul agent is sensitive to a HUP during its initialization process, such as when invoked via 'knife ssh'.  Adding a nohup to the invocation or adding a sleep after it seems to clear up the issue.

I've opened an issue against consul itself as well.

jhmartin:
Give consul time to install its HUP signal handler to avoid immediate shutdown when invoked via knife ssh, fixes #125.

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,14 @@ Installs and configures [Consul][1] client, server and UI.
     </td>
     <td><tt>nil</tt></td>
   </tr>
+  <tr>
+    <td><tt>['consul']['startup_sleep']</tt></td>
+    <td>Integer</td>
+    <td>
+        Delay to return-of-control when invoked via an init.d script. Protects consul from an early HUP.
+    </td>
+    <td><tt>3</tt></td>
+  </tr>
 </table>
 
 ### Databag Attributes (optional)
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -138,3 +138,6 @@ default['consul']['extra_params'] = {}
 default['consul']['atlas_autojoin'] = false
 default['consul']['atlas_cluster'] = nil
 default['consul']['atlas_token'] = nil
+
+# Init script startup delay
+default['consul']['startup_sleep'] = 3
--- a/recipes/service.rb
+++ b/recipes/service.rb
@@ -224,7 +224,8 @@ when 'init'
     source init_tmpl
     mode init_mode
     variables(
-      consul_logfile: node['consul']['logfile']
+      consul_logfile: node['consul']['logfile'],
+      startup_sleep: node['consul']['startup_sleep']
     )
     notifies :restart, 'service[consul]', :immediately
   end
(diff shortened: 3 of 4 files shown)


tgwizard:
Should this really be part of the init script and not the chef setup process?

jhmartin:
@tgwizard See https://github.com/hashicorp/consul/issues/677 ; the issue occurs when called outside of chef.  It just so happens that `knife ssh ... /etc/init.d/consul start` also triggers the behavior.

tgwizard:
@jhmartin ah, ok. Still, this will affect starts and stops when not running through knife ssh etc., slowing everything down. Wouldn't

```
knife ssh ... "/etc/init.d/consul start && sleep 3"
```

work for you?

jhmartin:
@tgwizard I suppose it would, but that means other users also have to know about this behavior and think about each invocation mechanism (Rundeck, mcollective, ssh-for-loop, fabric, ansible, etc) and determine if it requires a timing-safe wrapper. Startup scripts do not normally have that kind of environmental sensitivity, so having consul allow it violates Least Surprise.

dpkp:
I think this needs to be in the init script.  You can trigger this bug simply by writing a recipe that does `notifies :reload, 'service[consul]'` -- i.e., what the README does in sample code.  If some other resource has notified a restart then you'll hit this (this happened to us).

Might I suggest checking `consul info` in a loop instead of sleeping 3 secs?  For the upstart template you can do this via:

```
# Fix for upstream issue. see https://github.com/johnbellone/consul-cookbook/issues/125
post-start script
  echo "Sleep a bit to allow consul time to install signal handlers"
  while ! consul info; do sleep 1; done
end script
```

[do you mind adding this patch ^^^ for `templates/default/consul.conf.erb` ?  Dont think it makes sense for me to do a separate PR for upstart here.]

dpkp:
further testing and it seems like we can't rely on `consul info` as a signal that startup is done.  sleep and pray may be the best answer here for now.

johnbellone:
I am not a huge fan of this. Is it really what the Consul folks suggest?

dpkp:
Is there an alternate fix for this? afaik, hashicorp currently recommends adding sleep to startup command: https://github.com/hashicorp/consul/issues/677#issuecomment-73429991

tgwizard:
If we add it, perhaps it should be customizable? Not everyone will have the use-case of starting consul over ssh like this.

johnbellone:
I am willing to put these changes in if someone wants to submit a PR. I'm going to have one more release prior to merging in #126.

jhmartin:
I've updated this PR against current master and made the sleep configurable, but defaults to a safe value. The value can be set to 0 if the environment does not need the extra safety.

@dpkp I'm not familar enough w/upstart -- you can submit a PR against my branch though and I'll include it, or you can submit a separate PR.

PR #135, opened 2015-02-24
Add support for Consul 0.5.0 and Atlas auto-join

shanesveller:
Commit adfa120174c85a3c61d6ab6be9ca390f3af05af2 can probably be ignored from mainline. I took a glance at rebasing this off of `develop` branch,
but it seems like it's in the middle of a sizeable rewrite.

Please keep or discard as much as you the maintainers find useful, in light of #126 etc.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -49,3 +49,11 @@ suites:
       consul:
         service_mode: cluster
         bootstrap_expect: 1
+  - name: atlas
+    run_list:
+      - recipe[consul::default]
+    attributes:
+      consul:
+        atlas_autojoin: true
+        atlas_cluster: <%= ENV.fetch('ATLAS_CLUSTER', 'example/cluster') %>
+        atlas_token: <%= ENV.fetch('ATLAS_TOKEN', 'NOT_REAL') %>
--- a/README.md
+++ b/README.md
@@ -256,6 +256,30 @@ Installs and configures [Consul][1] client, server and UI.
     <td>This provides the address of a statsd instance (UDP).</td>
     <td><tt>nil</tt></td>
   </tr>
+  <tr>
+    <td><tt>['consul']['atlas_autojoin']</tt></td>
+    <td>Boolean</td>
+    <td>
+        Determines whether Consul attempts to auto-join the cluster provided by <tt>atlas_cluster</tt> using the value of <tt>atlas_token</tt>
+    </td>
+    <td><tt>false</tt></td>
+  </tr>
+  <tr>
+    <td><tt>['consul']['atlas_cluster']</tt></td>
+    <td>String</td>
+    <td>
+        Name of Atlas cluster to auto-join
+    </td>
+    <td><tt>nil</tt></td>
+  </tr>
+  <tr>
+    <td><tt>['consul']['atlas_token']</tt></td>
+    <td>String</td>
+    <td>
+        API token used for Atlas integration
+    </td>
+    <td><tt>nil</tt></td>
+  </tr>
 </table>
 
 ### Databag Attributes (optional)
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -118,3 +118,8 @@ default['consul']['client_interface'] = nil
 default['consul']['client_addr']  = '0.0.0.0'
 default['consul']['serve_ui']     = false
 default['consul']['extra_params'] = {}
+
+# Atlas support
+default['consul']['atlas_autojoin'] = false
+default['consul']['atlas_cluster'] = nil
+default['consul']['atlas_token'] = nil
--- a/recipes/_service.rb
+++ b/recipes/_service.rb
@@ -177,6 +177,17 @@ if node.consul.verify_incoming || node.consul.verify_outgoing
   end
 end
 
+# Atlas integration
+if node.consul.atlas_autojoin or node.consul.atlas_token
+  cluster = node.consul.atlas_cluster
+  token = node.consul.atlas_token
+  raise "atlas_cluster is empty or nil" if cluster.empty? or cluster.nil?
+  raise "atlas_token is empty or nil" if token.empty? or token.nil?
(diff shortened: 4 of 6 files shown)


.kitchen.yml
```diff
@@ -53,3 +53,16 @@ suites:
...
+  - name: atlas
+    run_list:
+      - recipe[consul::default]
+    attributes:
+      consul:
+        atlas_autojoin: true
+        atlas_cluster: <%= ENV.fetch('ATLAS_CLUSTER', 'example/cluster') %>
+        atlas_token: <%= ENV.fetch('ATLAS_TOKEN', 'NOT_REAL') %>
+    excludes:
+      - centos-7.0
```
johnbellone: Any reason why EL won't work, or  you just haven't tested it?

shanesveller: If I recall, I believe CentOS 5.x and 7.x were stumbling on the libarchive dependency when I tried, perhaps due to package naming. They seemed to fail similarly on master without my changes so I haven't started digging yet.

donaldguy:
Ideally this should support auto-bootstrapping the cluster with `bootstrap_expect`; but presently the logic in `_service.rb` elides that.

I have prototyped that here: https://github.com/donaldguy/consul-cookbook/commit/7cf273aaffc220af6dc42cd74ddb76baba168015

It worked for me, starting from an up and associated with Atlas but unclustered set of servers, bringing them all down, deleteing the contents of the data dir (probably could just done `/raft` and/or `/serf`) and bringing them up one at a time.

.kitchen.yml
logankoester: @shanesveller libarchive chooses the package name with a [switch on `platform_family?`](https://github.com/reset/libarchive-cookbook/blob/master/recipes/default.rb#L17).

I had the same problem on Arch Linux, and all it took was [an extra case](https://github.com/logankoester/libarchive-cookbook/blob/arch/recipes/default.rb#L28) to resolve it.

Guessing CentOS is in a similar position.

https://github.com/reset/libarchive-cookbook/pull/6

johnbellone:
@shanesveller Please rebase!

shanesveller:
@johnbellone I've rebased my branch, but it looks like CentOS Kitchen tests are currently failing - maybe on master too. Will investigate more tomorrow.

donaldguy:
Also please make sure something with `bootstrap_expect` support lands; I also really don't think you want the `retry_join` or `start_join` server lists if you are working with `atlas_join`

I have all that on the branch(es) here but I haven't tried the tests: https://github.com/donaldguy/consul-cookbook/commit/7cf273aaffc220af6dc42cd74ddb76baba168015

shanesveller:
@johnbellone Rebased one more time - CentOS tests fail on master too, and my company doesn't use it so I'm afraid I can't really invest any time digging further.

@donaldguy I'm not clear on what your concerns are, can you elaborate here in the comments? My company tested this code as-is with an EC2 auto-scale group created from Packer-built AMIs and had no trouble scaling the cluster from 0 to 3 and up-and-down from there. The [Atlas guide](https://www.consul.io/docs/guides/atlas.html) and [Configure documentation](https://www.consul.io/docs/agent/options.html) don't seem to mention `retry_join`/`start_join` in conjunction with Atlas at all.

donaldguy:
So upon further review, this is a case of user error, but I suspect I won't be the only one. 

the TL;DR is **I didn't use the `service_mode` `cluster` cause I wasn't clear I was supposed to**, because I didn't see the Getting Started section all the way at the bottom of the README (I saw and didn't want to use the chef-provisioning script under Usage, and I didn't think to look for more setup instructions after all the LWRP examples). 

Maybe the fix is just putting it higher in the README or linking to it under Usage. It may also be making the `cluster` and `server` values synonymous, cause I am unconvinced `server` in its current incarnation is ever useful. @johnbellone please advise on the dispensation of `server`

---
While I noticed `cluster` as an value checked for `service_mode` in `_service.rb`, because there was nothing in the Consul docs about a `cluster` mode, I assumed it was just for the chef-provisioning recipe and/or something legacy. I understood the one-only semantics of `bootstrap`, so `server` struck me as the correct value for that attribute

But what happens with this value and Atlas is confusing, when I first spun up, `consul members` showed my three servers and things looked good, but doing `curl localhost:8500/v1/status/leader` returned `""` and the Atlas consul UI would not update (In an email exchange with Hashicorp folks, they say they'll probably add a no leader warning to explain that situation. Cause I was just seeing "Your infrastructure is healthy; 0 nodes".)

So the problems were/are three fold:
  1. I wasn't initially aware of the necessity `bootstrap_expect` parameter; this is on me.
  2. Using `service_mode` of `server` does not populate either the default nor a custom `bootstrap_expect` parameter through to the underlying config, causing no autoclustering to ever occur. I am not sure why this would ever be desirable.
 3. There is, however, written still a `start_join` or 'retry_join` block in the config. Since I didn't use the chef-provisioning script or manually populate `servers` this is blank. This seems minimally aesthetically redundant when using the Atlas integration, and at worst having either a blank or outdated values here could prevent autoclustering from working in the event of a leader failure. I'd need to do more testing to determine which conclusively

---
so you can probably safely ignore my branch and perhaps move the `cluster` v `server` discussion to a different PR/issue, but I will still suggest that, since Atlas seems to make the `retry_join`/`start_join` configs unnecessary that perhaps lines of the form `service_config[join_mode] = node['consul']['servers']` should be `service_config[join_mode] = node['consul']['servers'] unless node.consul.atlas_autojoin`

jjhuff:
Hey, right now, `atlas_cluster` is required, but consul doesn't seem to require it.

It seems that auto-join and the SCADA/infrastructure stuff are two separate things (both require `atlas_token`).  It'd be nice to be able to enable auto_join without SCADA.

johnbellone:
#126 is about to land on master. Are these attributes exposed via configuration?

PR #142, opened 2015-03-11
Make service enable conditional on an attribute

donaldguy:
In some workflows (e.g. 'image baking') you may want a chef run that
installs consul, but to delay the actual start until a later run

I only tested the runit path (and that on the [consul-0.5.0-conditional](https://github.com/donaldguy/consul-cookbook/tree/consul-0.5.0-conditional) branch but thought I'd get eyes on it anyway


johnbellone:
I am closing this out because of impeding #126 landing on master. You'll be able to write a recipe and have the service "enabled" but not "started".

PR #161, opened 2015-04-06
Refactor start and stop timeouts in consul-init

jon918:
Based on the mysql-server init script.

I found that the 10 second timeout on shutdown wasn't enough, I was getting some errors that were resolved by increasing to 30.


mjuszczak:
:+1:

PR #163, opened 2015-04-16
Adding rejoin after leave for auto cluster remediation.

arodd:
I would like to be able to enable this value from within my chef-attributes.  Based on a quick code analysis it appears I just need to add the name of the parameter in this array.


arodd:
Maybe it is just that the "extra_params" attribute isn't well documented but I was able to add my extra configs in there for now.  What determination is being used to decide which properties end up in extra params or in this array that essentially does the exact same thing?

PR #188, opened 2015-06-23
Adding GOMAXPROCS support for systemd.

joshgarnett:
When using systemd it currently always uses the default GOMAXPROCS of 1.  This change follows the documentation at http://fedoraproject.org/wiki/Packaging%3aSystemd#EnvironmentFiles_and_support_for_.2Fetc.2Fsysconfig_files

### DIFF
--- a/templates/default/consul-systemd.erb
+++ b/templates/default/consul-systemd.erb
@@ -6,12 +6,13 @@ After=basic.target network.target
 [Service]
 User=<%= node['consul']['service_user'] %>
 Group=<%= node['consul']['service_group'] %>
+EnvironmentFile=-<%= node['consul']['etc_config_dir'] %>
 ExecStart=<%= Chef::Consul.active_binary(node) %> agent \
   -config-dir <%= node['consul']['config_dir'] %>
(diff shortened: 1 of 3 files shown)


templates/default/consul-systemd.erb
```diff
@@ -6,12 +6,13 @@ After=basic.target network.target
...
 User=<%= node['consul']['service_user'] %>
 Group=<%= node['consul']['service_group'] %>
+EnvironmentFile=-<%= node['consul']['etc_config_dir'] %>
 ExecStart=<%= Chef::Consul.active_binary(node) %> agent \
   -config-dir <%= node['consul']['config_dir'] %>
 ExecReload=/bin/kill -HUP $MAINPID
 KillMode=process
 Restart=on-failure
-RestartSec=42s
+RestartSec=15s
```
johnbellone: What's the reason for this change?

joshgarnett: 42 seemed like an arbitrary number (bring a towel, meaning of life, etc...).  15s is more in line with what I'd expect for a recovery time.

PR #189, opened 2015-07-02
use node array to refer to variables

jedineeper:
I use this cookbook via the "include_recipe 'consul'" method as I include this in one of my "base-server" cookbooks.

When I try to pass atlas autojoin/cluster/token variables through my cookbook the variables are not picked up by consul. I had a look through the code and found that the variables were referenced via "node.consul.atlas_autojoin" rather than "node['consul']['atlas_autojoin']". I'm not sure if there was a particular reason for this but changing it like this means the variables are picked up correctly.

@shanesveller originally wrote this in https://github.com/johnbellone/consul-cookbook/pull/135 not sure if there was a reason you referenced the variables this way Shane so happy for commentary if there was a particular reason :)

Tests still pass correctly, though I'm not sure how to word a test to cover this particular case?

Thanks,
Craig

### DIFF
--- a/recipes/service.rb
+++ b/recipes/service.rb
@@ -180,13 +180,13 @@ if node.consul.verify_incoming || node.consul.verify_outgoing
 end
 
 # Atlas integration
-if node.consul.atlas_autojoin or node.consul.atlas_token
-  cluster = node.consul.atlas_cluster
-  token = node.consul.atlas_token
+if node['consul']['atlas_autojoin'] or node['consul']['atlas_token']
+  cluster = node['consul']['atlas_cluster']
+  token = node['consul']['atlas_token']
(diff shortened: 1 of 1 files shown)


shanesveller:
Square-brackets notation and dot notation for attributes is a stylistic choice and should behave mostly identically in either case. If you're seeing otherwise, that sounds like a bug. It's been some months since I wrote that code, but I actually do prefer square-brackets notation in my personal work. Odds are I was matching other code in this cookbook due to a [FoodCritic rule](http://acrmp.github.io/foodcritic/#FC019) or something of that nature - I can't think of why else I would write it this way.

jedineeper:
rebased on master 5 July 2015.

PR #196, opened 2015-07-16
Windows: resolved merge conflicts and added tests

gdavison:
This adds support for installing on Windows using the `install_method` of `windows`. It uses the Chocolatey package manager.

I have made the Windows tests conditional on setting the `TEST_WINDOWS` environment variable to a truthy value. I tried using vagrant boxes from [https://github.com/joefitzgerald/packer-windows] and [https://github.com/boxcutter/windows],  but I was not able to make them work. The box used is approximately 3 GB to download. Provisioning only seems to work if you login at boot time; annoying, but it works.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -10,6 +10,13 @@ platforms:
   - name: ubuntu-12.04
   - name: centos-7.0
   - name: centos-6.5
+  <% if ENV['TEST_WINDOWS'] %>
+  - name: windows-2012r2-wrock
+    driver_config:
+      box_url: https://wrock.ca.tier3.io/win2012r2-virtualbox.box
+      customize:
+        usbehci: "off"
+  <% end %>
 
 suites:
   - name: default
@@ -70,3 +77,14 @@ suites:
         atlas_autojoin: true
         atlas_cluster: <%= ENV.fetch('ATLAS_CLUSTER', 'example/cluster') %>
         atlas_token: <%= ENV.fetch('ATLAS_TOKEN', 'NOT_REAL') %>
+  <% if ENV['TEST_WINDOWS'] %>
+  - name: windows
+    includes:
+      - windows-2012r2-wrock
+    run_list:
+      - recipe[consul::default]
+    attributes:
+      consul:
+        install_method: windows
+        service_user: vagrant
+  <% end %>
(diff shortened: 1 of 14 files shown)


gdavison:
The box I was using appears to have been taken offline today. I've changed to a new box from the same source. You no longer need to log in, but kitchen seems to try on first run to use ssh to connect. I cancel the first run with ^C, then re-run the test, and it uses winrm. No idea why.

johnbellone:
What's the best way to test this?

gdavison:
I've fixed some of the Windows VM stuff by manually setting some values that I thought Kitchen was supposed to set. I'm verifying that all tests pass right now, and will re-push when that's done.

gdavison:
It's all updated. You can run `TEST_WINDOWS=1 kitchen test "^[^rs].*(ub|win)"`. It skips the `centos`, `runit`, and `source` tests, and runs the `windows` test.

johnbellone:
This is failing for me. Are you using ChefDK?

gdavison:
Yes, ChefDK 0.6.2 on Mac OS 10.10. What errors are you getting?

.kitchen.yml
```diff
@@ -10,9 +10,22 @@ platforms:
   - name: ubuntu-12.04
   - name: centos-7.0
   - name: centos-6.5
+  <% if ENV['TEST_WINDOWS'] %>
```
johnbellone: I do not see a reason why we need to keep the Windows test support on an environment variable? I've removed it locally and all looks good to me. Please remove this bit.

.kitchen.yml
```diff
@@ -63,10 +85,23 @@ suites:
...
     run_list:
       - recipe[consul::default]
     attributes:
       consul:
         atlas_autojoin: true
         atlas_cluster: <%= ENV.fetch('ATLAS_CLUSTER', 'example/cluster') %>
-        atlas_token: <%= ENV.fetch('ATLAS_TOKEN', 'NOT_REAL') %>
\ No newline at end of file
+        atlas_token: <%= ENV.fetch('ATLAS_TOKEN', 'NOT_REAL') %>
+  <% if ENV['TEST_WINDOWS'] %>
```
johnbellone: The same goes here as well.

PR #199, opened 2015-07-23
Adding open files configuration for Upstart.

darron:
In our environment, we're getting to the place where Consul doesn't have enough open files by default.

Upstart doesn't respect `/etc/security/limits.conf`: https://bugs.launchpad.net/ubuntu/+source/upstart/+bug/938669

So you have to add it manually to the `/etc/init/consul.conf`: http://upstart.ubuntu.com/wiki/Stanzas#limit

This seemed like the best way to do it given the cookbook creates those files.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -105,6 +105,10 @@ default['consul']['ports'] = {
   "server"   => 8300,
 }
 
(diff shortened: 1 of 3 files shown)


tgwizard:
:+1:

darron:
Thanks!

PR #203, opened 2015-07-26
resource and library to handle acls

InformatiQ:
A lib to manage acls on a consul server using the HTTP api and the resources that use it.

The plan is to later add more resources/providers that use that same consul_client.rb lib instead of writing files.


johnbellone:
@InformatiQ Please take a look at the new libraries which were merged into #126 and refactor to follow this form.

PR #208, opened 2015-08-20
Remove opinionated lock on golang

scalp42:
Hi @johnbellone,

The lock on `golang` is too opinionated, folks should be able to lock themselves (or default to unlocked) in my opinion.

Thanks in advance


tgwizard:
:+1:

scalp42:
@johnbellone any chance to get a patch bump for this?

PR #220, opened 2015-09-22
Add provider for consul acls

kamaradclimber:
This patch introduces acl provider for consul.
Most consul resources can injected as configuration file, acls however
have to interacted with through http api.
ConsulClient library exposes minimum helpers for such interaction.

This patch is based on the work of @InformatiQ

### DIFF
--- a/libraries/consul_acl.rb
+++ b/libraries/consul_acl.rb
@@ -1,0 +1,65 @@
+# Cookbook: consul
+# License: Apache 2.0
+#
+# Copyright (C) 2014, 2015 Bloomberg Finance L.P.
+#
+require 'poise'
+
+module ConsulCookbook
+  module Resource
+    # @since 1.0.0
+    class ConsulACL < Chef::Resource
+      include Poise(fused: true)
+      provides(:consul_acl)
+      default_action(:create)
+
+      # @!attribute name
+      # @return [String]
+      attribute(:name, name_attribute: true, kind_of: String)
+      # @!attribute rules
+      # @return [String]
+      attribute(:rules, kind_of: String)
+      # @!attribute id
+      # @return [String]
+      attribute(:id, kind_of: String)
+      # @!attribute type
+      # @return [String]
+      attribute(:type, kind_of: String, default: 'client', required: true)
+
+      action :create do
+        ruby_block @new_resource.name do
+          data = {}
+          data['Name'] = new_resource.name
+          data['Type'] = new_resource.type
+          data['Rules'] = new_resource.rules
+          data['ID'] = new_resource.id unless new_resource.id.nil?
+          block { client.manage_acl(data) }
+          only_if { !current_resource || update_required? }
+        end
+      end
+
+      action :delete do
+        ruby_block new_resource.name do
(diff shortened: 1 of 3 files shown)


kamaradclimber:
@johnbellone is there anything I can improve to get this merged ?

kamaradclimber:
@johnbellone could you give me some feedback on this change ?

johnbellone:
@kamaradclimber Apologies for not looking at this sooner.

I am wondering if we should be using an existing Ruby library for the Consul HTTP client? There seem to be a few on Rubygems.org already.

libraries/consul_acl.rb
```diff
@@ -0,0 +1,65 @@
...
+          block { client.delete_acl(new_resource.name) }
+          only_if { current_resource }
+        end
+      end
+
+      def client
+        @client ||= Chef::Consul::Client.new(nil, node['consul']['ports']['http'], node['consul']['acl_master_token'])
+      end
+
+      def load_current_resource
```
johnbellone: Does this actually work? I have not had a change to take a look at the Poise fused resources more deeply, but from what I recall all defined methods are on the resource not the provider. This may be a case where you need a separate provider, or you'll have to explicitly load the resource in the action.

libraries/consul_acl.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      action :delete do
+        ruby_block new_resource.name do
+          block { client.delete_acl(new_resource.name) }
+          only_if { current_resource }
+        end
+      end
+
+      def client
+        @client ||= Chef::Consul::Client.new(nil, node['consul']['ports']['http'], node['consul']['acl_master_token'])
```
johnbellone: Please pass in the node attributes as properties (attributes) on the resource. This allows us to validate at convergence time and also makes definition of resource more explicit.

libraries/consul_acl.rb
```diff
@@ -0,0 +1,65 @@
...
+  module Resource
+    # @since 1.0.0
+    class ConsulACL < Chef::Resource
+      include Poise(fused: true)
+      provides(:consul_acl)
+      default_action(:create)
+
+      # @!attribute name
+      # @return [String]
+      attribute(:name, name_attribute: true, kind_of: String)
```
johnbellone: This is redundant. The name attribute is by default on all resources.

johnbellone:
I am also going to need more tests here. I took a look at the code and am a little wary of executing arbitrary ruby blocks. Could you please explain what you're trying to do there? Maybe we could work on refactoring it.

kamaradclimber:
thanks for you feedback. We still use the pre1.0 version of this cookbook in my organisation so I relied on rspec to validate the patch (version 1.0 does not work on windows 2012R2 probably due to a chef bug)

I'll have a look at your comments in the next days.

kamaradclimber:
I can answer about using them gem: @InformatiQ has done the PR against the diplomat gem (https://github.com/WeAreFarmGeek/diplomat/pull/46) that contains the code.

since the diplomat PR needs to be improved, let's put this one on hold until we are able to directly use diplomat gem.

PR #223, opened 2015-09-25
Skip SElinux recipe on non linux OSes.

kamaradclimber:


kamaradclimber:
thanks for merging this, I am not sure it is sufficient to fix the windows setup so I may provide other patches in this direction.

PR #231, opened 2015-10-22
Install recipe

NickLaMuro:
Description
---------------
For use cases like `packer` or other OS artifact builders, it is helpful to be able to only install the software that is need without configuring it to run as a service on initial boot.  For example, you can install consul on a instance and put in place a few standard checks that would be on all machines (Memory, CPU, Disk Utilization) and pack that into your build and it is preinstalled when you boot up your box and the provisioning step only needs configure any additional customizations.

I was originally going to create a `consul::install` recipe, but based on how baked in the installation code is in the service provider, it was not really possible.  This ends up making this a bit of a hack because the recipe ends up adding a service only to delete it, but based on what I can tell, deleting it will only remove the base config file and the service definition, but leave the binary in place.

My original workaround without this change was going to be the following:

```
include_recipe 'consul'

consul_service node['consul']['service_name'] do
  action :disable
end
```

But the delayed `subscribes :restart` that is baked into the `consul::default` recipe will actually fail because the service definition file on the machine will no longer exist.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -31,6 +31,11 @@ default['consul']['service']['config_dir'] = '/etc/consul'
 default['consul']['service']['binary_url'] = "https://dl.bintray.com/mitchellh/consul/%{filename}.zip" # rubocop:disable Style/StringLiterals
 default['consul']['service']['source_url'] = 'https://github.com/hashicorp/consul'
 
+default['consul']['install']['install_method'] = 'binary'
+default['consul']['install']['config_dir'] = '/etc/consul'
+default['consul']['install']['binary_url'] = "https://dl.bintray.com/mitchellh/consul/%{filename}.zip" # rubocop:disable Style/StringLiterals
+default['consul']['install']['source_url'] = 'https://github.com/hashicorp/consul'
+
 default['consul']['version'] = '0.5.2'
 default['consul']['checksums'] = {
   '0.5.0_darwin_amd64' => '24d9758c873e9124e0ce266f118078f87ba8d8363ab16c2e59a3cd197b77e964',
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -26,11 +26,6 @@ default['consul']['config']['ports'] = {
   'server'   => 8300
 }
 
-default['consul']['service']['install_method'] = 'binary'
-default['consul']['service']['config_dir'] = '/etc/consul'
-default['consul']['service']['binary_url'] = "https://dl.bintray.com/mitchellh/consul/%{filename}.zip" # rubocop:disable Style/StringLiterals
-default['consul']['service']['source_url'] = 'https://github.com/hashicorp/consul'
-
 default['consul']['install']['install_method'] = 'binary'
 default['consul']['install']['config_dir'] = '/etc/consul'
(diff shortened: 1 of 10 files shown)


johnbellone:
I would prefer having a separate recipe for doing exactly this. What do you think?

It would also be great if you could add a section on embedding this into a Packer built image.

NickLaMuro:
@johnbellone I think that makes sense, I just didn't want to get too invasive on my first run of things.  I can do a spike for this when I get a chance and you can let me know you think.

NickLaMuro:
@johnbellone Alright, here is the spike as promised.

Rational
-----------
So, I ended up making a completely new lib strictly for installing consul, which does mostly what was done in `consul_service`, with a few tweaks:

* Doesn't deal with making a poise service (probably obvious...)
* Skips package resources on windows platforms since we are working with chocolatey to install the package on that platform
* Handles uninstall cases a little more verbosely

I did this mainly because the `consul_service` was doing more than setting up a service, but also installing consul, and it was too hard to write the `consul::install` recipe with `consul_service` being the only way of installing things in this cookbook.  This unfortunately lead to a massive addition to the repo, and because I am a bit of an "over tester", quite the verbose set of chefspec tests for that resource.


I am going to try fixing the broken tests I now have on travis (working on my machine of course...), and breaking out installation code in the `consul_service` and updating the default recipe to make use of `consul_install`, and also make sure the integration tests are working as expected.  I think most of what I will be doing will a little more obvious now that I have the ground work laid out, and you can tell me what you would rather have done differently if needed.

johnbellone:
I am wondering if this would make more sense to simply be an install action on the service provider. There would be a separate recipe called install (like you have already). The enable action would call action_install.

NickLaMuro:
That would definitely make this a smaller change :smile:

I have just getting familiar with the poise libs this past week, so I was mostly trying to stay within the actions provided by `poise-service` out of the box, which install isn't one of them.  That said, what you are saying definitely seems possible and a much nicer for you to deal with, so I will try looking into that.  If it is pretty easy to do and doesn't require too much hackary, I will probably end up scraping this branch since it will just turn into a much of reverts.

johnbellone:
That's a good point. I think I remember @coderanger telling me not to add actions to poise-service..

coderanger:
The optimal way to handle this would probably be separate providers for `consul_service` for each install type.

NickLaMuro:
Alright, I think this is where I want it... or at least for now.

A few things about the failing tests:

* I Took out the windows/chocolatey tests since I wasn't sure why they were failing on `.travis` but not my machine.
* Not sure what is up with the ruby 2.2 build, but I think that has been failing on master as well
* The kitchen tests are passing for the binary install at least.  For the failures, here is what I ran into:
  * `poise-service` seems to detect `upstart` as the provider for `centos-6.6`, but 'serverspec` is looking with `chkconfig` to determine if the process is running.
  * The source builds didn't work because the package manager wasn't up to date
  * The Package builds failed because it couldn't find a `consul` package
  * For all of these, I was unsure if this was due to my changes or if they have always been failing, and haven't attempted any solutions to fix them since I am unsure where the integration tests fall in the pipeline.


Should I look into splitting up `consul_install` into separate providers?

Issue #242, opened 2015-11-24
Basis for selinux set to disabled


ronlipke:
What is the basis for setting selinux to disabled in this cookbook?  I see it changed from permissive sometime between version 1.2.0 and 1.3.0.
We have some compliance restrictions that require selinux and was surprised to see it outright disabled here.
Thanks!

mdavis7771:
+1

agperson:
But why is this cookbook managing SELinux at all?  I'm confused as to why it is in this purview at all.

mdavis7771:
I agree with @agperson. As a security professional, having no documentation of this change and adding it to the run list could dramatically impact my security. SELinux should be managed be a separate wrapper cookbook.

johnbellone:
@mdavis7771 It is a long story, but ultimately it is a relic from an incorrect generation of this cookbook about a year ago. I do not have a problem removing this cookbook touching SELinux at all.

johnbellone:
I'll accept a PR here if either @mdavis7771 or @agperson is willing to do so. We'll need to increment the minor release.

mdavis7771:
@johnbellone, Understood, I'll do a PR later this week.

Ginja:
#256 should fulfill this.

PR #243, opened 2015-11-30
Improve selinux support

Ginja:
Added support for choosing selinux state

Using the selinux state attribute (from the selinux cookbook) to specify what state selinux should be in. Also changed the default state to 'permissive'. Addresses #242.


attributes/default.rb
```diff
@@ -51,3 +51,9 @@
   '0.5.2_windows_386'  => '2e866812de16f1a6138a0fd1eebc76143f1314826e3b52597a55ac510ae94be6',
   '0.5.2_web_ui'       => 'ad883aa52e1c0136ab1492bbcedad1210235f26d59719fb6de3ef6464f1ff3b1'
 }
+
+###
+# selinux
+###
+
+default['selinux']['state'] = 'permissive'
```
johnbellone: I would not put a different cookbook's attribute here as it'll take awhile for someone to track this down. If we want to default the attribute value we should do it prior to the include in the default recipe so that it is explicit. 
```ruby
if node['os'] == 'linux'
  node.default['selinux']['state'] = 'permissive'
  include_recipe 'selinux::default'
end
```

Ginja:
@johnbellone good suggestion. I've made the change.

PR #249, opened 2015-12-07
Add a version to poise-boiler to fix Travis

elyscape:


elyscape:
I have absolutely no idea why this drops down the code coverage.

PR #252, opened 2015-12-10
Updated Consul Download URLs

Ginja:
Looks like consul is now prefixing their download URLs with 'consul'. Updated appropriate attributes, and binary_filename methods.

Reference - https://releases.hashicorp.com/consul/0.6.0/

PR #253, opened 2015-12-11
Update URL for UI package

bdclark:
A PR was recently merged to update the URL of the main Consul binary; this PR updates the consul_ui resource to default to the new URL.

### DIFF
--- a/libraries/consul_ui.rb
+++ b/libraries/consul_ui.rb
@@ -26,7 +26,7 @@ module ConsulCookbook
 
       # @!attribute binary_url
       # @return [String]
-      attribute(:binary_url, kind_of: String, default: 'https://releases.hashicorp.com/consul/%{version}/%{filename}.zip')
+      attribute(:binary_url, kind_of: String, default: 'https://releases.hashicorp.com/consul/%{version}/consul_%{filename}.zip')
 
       # @!attribute source_url
       # @return [String]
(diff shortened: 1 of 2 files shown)


libraries/consul_ui.rb
```diff
@@ -26,7 +26,7 @@ class ConsulUI < Chef::Resource
 
       # @!attribute binary_url
       # @return [String]
-      attribute(:binary_url, kind_of: String, default: 'https://releases.hashicorp.com/consul/%{version}/%{filename}.zip')
+      attribute(:binary_url, kind_of: String, default: 'https://releases.hashicorp.com/consul/%{version}/consul_%{filename}.zip')
```
Ginja: I'm pretty sure merging this PR in combination with #252  will break the cookbook. The default value for :binary_url will return `consul_consul_0.6.0_webui.zip` and that URL doesn't exist. I'd suggest reverting this merge.

bdclark:
I believe @Ginja is right.  I submitted this PR in the spirit of #250 already being merged... at the time #252 was still outstanding.  Since each took a different approach to solving the URL issue, it looks like merging both may have broke the world.

libraries/consul_ui.rb
jslusher: This PR has broken the cookbook for me. 

```Net::HTTPServerException: remote_file[https://releases.hashicorp.com/consul/0.5.2/consul_consul_0.5.2_web_ui.zip```

Ginja: @johnbellone Could we merge #256 ?

PR #254, opened 2015-12-11
Guard against owning /etc

bdclark:
The `consul_config` resource ensures the directory exists for the config file and is owned by the service user.  Unfortunately since the default path for the config file is `/etc/config.json`, this cookbook is changing ownership of `/etc` to the Consul user, which is causing problems on some of my nodes.

This PR simply guards the directory resource from managing the directory if it is `/etc`. This may not be the best approach, but it was the simplest way to solve the problem.  Please let me know if there's a better way and I'll submit another PR. Thanks!

### DIFF
--- a/libraries/consul_config.rb
+++ b/libraries/consul_config.rb
@@ -149,6 +149,7 @@ module ConsulCookbook
             owner new_resource.owner
             group new_resource.group
             mode '0755'
+            not_if { ::File.dirname(new_resource.path) == '/etc' }
           end
 
           file new_resource.path do


Ginja:
@johnbellone, any reason to not change the default values for `node['consul']['config']['path']` & `node['consul']['config']['data_dir']` to:

```
node.default['consul']['config']['path'] = '/etc/consul/consul.json'
node.default['consul']['config']['data_dir'] = '/etc/consul/data'
```
?

johnbellone:
I don't have any complaints here.

johnbellone:
@Ginja do you want to put a separate PR in?

bdclark:
@Ginja would that be changing the defaults for `node.default['consul']['config']['data_dir']` or `node.default['consul']['config']['config_dir']`?  Not sure if data_dir is relevant to the issue, but maybe I'm mistaken.

Ginja:
I'd say both as I like to keep everything Consul in one location, but if anyone has any objections to that the default location for `node.default['consul']['config']['data_dir']` (i.e. `/var/lib/consul`) is fine.

bdclark:
Before taking the easy way out and guarding `/etc` I was going to suggest...
```
node.default['consul']['config']['path'] = '/etc/consul/consul.json'
node.default['consul']['config']['config_dir'] = '/etc/consul/conf.d'
```
Leaving all things configuration in `/etc/consul/<foo>` and leaving data stuff in `/var/lib/consul`, however at the end of the day I'm with you and am fine with either.

johnbellone:
Now that I am reading this again I am a little concerned - the /etc/consul/data directory shouldn't be the default.

PR #257, opened 2015-12-13
Add ACL support

bdclark:
This is a rough draft of a resource to provide ACL support using the `diplomat` gem as an API client.  It's not ready to merge yet since there's no unit tests, etc... however I wanted to throw it out there for comments and suggestions.

Since most of the work is being handled by diplomat there's really not much to it. It's passing the included integration tests, is idempotent, notifies correctly, etc.

A few noteworthy points:
* Since it requires a chef_gem there needs to be a way to install it. Rather than doing anything fancy in the resource I decided to leave it up to the user to ensure the gem is installed first.  A `consul::client_gem` recipe is included with the cookbook to do the needful.
* The diplomat version is not pinned, but an attribute exists to set the version if desired.
* I'm requiring diplomat in each of the action blocks rather than the top of the resource to prevent the gem from needing to be installed at compile-time (is this best-practice?).
* There is no `:update`... just `:create` and `:delete`. After tinkering with the Consul API for awhile it looks like the only difference between create and update in Consul's API is the fact that create can be called without an ID (it'll create the ACL with a random UUID if ID is not provided). So, to keep things simple/predictable the create endpoint is used for create/update and an ID is required.

Comments?  Thanks!

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -46,3 +46,17 @@ suites:
     excludes:
       - centos-7.1
       - centos-6.6
+  - name: acl
+    run_list:
+      - recipe[consul::default]
+      - recipe[consul::client_gem]
+      - recipe[consul_spec::acl]
+    attributes:
+      consul:
+        config:
+          bootstrap: true
+          server: true
+          datacenter: fortmeade
+          acl_master_token: doublesecret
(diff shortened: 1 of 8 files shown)


johnbellone:
This looks good to me.

PR #259, opened 2015-12-30
Added Windows Support

### LINKED ISSUE #236: Windows install on 64-bit fails
When installing on Windows 64-bit, the install fails as there is no 64-bit version for Windows available.

```* consul_service[consul] action enable
  ================================================================================
  Error executing action `enable` on resource 'consul_service[consul]'
  ================================================================================

  KeyError
  --------
  key not found: "0.5.2_windows_amd64"

  Resource Declaration:
  ---------------------
  # In C:/chef/cache/cookbooks/consul/recipes/default.rb

   38: consul_service node['consul']['service_name'] do |r|
   39:   user node['consul']['service_user']
   40:   group node['consul']['service_group']
   41:   version node['consul']['version']
   42:   config_file config.path
   43:
   44:   node['consul']['service'].each_pair { |k, v| r.send(k, v) }
   45:   subscribes :restart, "consul_config[#{config.name}]", :delayed
   46: end

  Compiled Resource:
  ------------------
  # Declared in C:/chef/cache/cookbooks/consul/recipes/default.rb:38:in `from_file'

  consul_service("consul") do
    action [:enable]
    retries 0
    retry_delay 2
    default_guard_interpreter :default
    declared_type :consul_service
    cookbook_name "consul"
    recipe_name "default"
    user "consul"
    group "consul"
    version "0.5.2"
    config_file "/etc/consul.json"
    install_method "binary"
    config_dir "/etc/consul"
    binary_url "https://dl.bintray.com/mitchellh/consul/0.5.2_windows_amd64.zip"
    source_url "https://github.com/hashicorp/consul"
    package_name "consul"
    install_path "/srv"
  end


Running handlers:
[2015-11-17T11:34:23+02:00] ERROR: Running exception handlers
Running handlers complete
[2015-11-17T11:34:23+02:00] ERROR: Exception handlers complete
Chef Client failed. 2 resources updated in 26 seconds
[2015-11-17T11:34:23+02:00] FATAL: Stacktrace dumped to C:/chef/cache/chef-stacktrace.out
[2015-11-17T11:34:23+02:00] FATAL: KeyError: consul_service[consul] (consul::default line 38) had an error: KeyError: key not found: "0.5.2_windows_amd64"```

### PR DESCRIPTION
**Pull request by Ginja**
* Fixes #236
* Added Windows Support (binary install_method only) using NSSM
  to manage the service. NSSM runs consul as the local SYSTEM user.
  Support for running it as another user can/should be added later.
* Prevented the firewall cookbook from creating rules for disabled
  ports
* Added & Updated spec tests
* Disabled Style/ModuleFunction cop
* Simplified how attributes with paths for values are set
* Added version check to windows binary installation so that it can
  be upgraded

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -4,18 +4,23 @@
 #
 # Copyright 2014-2016, Bloomberg Finance L.P.
 #
+::Chef::Node.send(:include, ConsulCookbook::Helpers)
+
+# Only used on Linux
 default['consul']['service_name'] = 'consul'
 default['consul']['service_user'] = 'consul'
 default['consul']['service_group'] = 'consul'
 
-default['consul']['config']['path'] = '/etc/consul.json'
 default['consul']['config']['bag_name'] = 'secrets'
 default['consul']['config']['bag_item'] = 'consul'
-default['consul']['config']['data_dir'] = '/var/lib/consul'
-default['consul']['config']['ca_file'] = '/etc/consul/ssl/CA/ca.crt'
-default['consul']['config']['cert_file'] = '/etc/consul/ssl/certs/consul.crt'
+
+default['consul']['config']['path'] = join_path prefix_path, 'consul.json'
+default['consul']['config']['data_dir'] = join_path prefix_path, 'data'
+default['consul']['config']['ca_file'] = join_path prefix_path, 'ssl', 'CA', 'ca.crt'
+default['consul']['config']['cert_file'] = join_path prefix_path, 'ssl', 'certs', 'consul.crt'
+default['consul']['config']['key_file'] = join_path prefix_path, 'ssl', 'private', 'consul.key'
+
 default['consul']['config']['client_addr'] = '0.0.0.0'
-default['consul']['config']['key_file'] = '/etc/consul/ssl/private/consul.key'
 default['consul']['config']['ports'] = {
   'dns'      => 8600,
   'http'     => 8500,
@@ -27,8 +32,10 @@ default['consul']['config']['ports'] = {
 
 default['consul']['diplomat_version'] = nil
 
+default['consul']['service']['config_dir'] = join_path prefix_path, 'conf.d'
+default['consul']['service']['data_dir'] = join_path prefix_path, 'data'
+
 default['consul']['service']['install_method'] = 'binary'
-default['consul']['service']['config_dir'] = '/etc/consul'
 default['consul']['service']['binary_url'] = "https://releases.hashicorp.com/consul/%{version}/%{filename}.zip" # rubocop:disable Style/StringLiterals
 
 default['consul']['service']['source_url'] = 'https://github.com/hashicorp/consul'
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -14,11 +14,11 @@ default['consul']['service_group'] = 'consul'
 default['consul']['config']['bag_name'] = 'secrets'
 default['consul']['config']['bag_item'] = 'consul'
 
-default['consul']['config']['path'] = join_path prefix_path, 'consul.json'
-default['consul']['config']['data_dir'] = join_path prefix_path, 'data'
-default['consul']['config']['ca_file'] = join_path prefix_path, 'ssl', 'CA', 'ca.crt'
-default['consul']['config']['cert_file'] = join_path prefix_path, 'ssl', 'certs', 'consul.crt'
-default['consul']['config']['key_file'] = join_path prefix_path, 'ssl', 'private', 'consul.key'
+default['consul']['config']['path'] = join_path config_prefix_path, 'consul.json'
+default['consul']['config']['data_dir'] = join_path data_prefix_path, 'data'
+default['consul']['config']['ca_file'] = join_path config_prefix_path, 'ssl', 'CA', 'ca.crt'
+default['consul']['config']['cert_file'] = join_path config_prefix_path, 'ssl', 'certs', 'consul.crt'
+default['consul']['config']['key_file'] = join_path config_prefix_path, 'ssl', 'private', 'consul.key'
 
 default['consul']['config']['client_addr'] = '0.0.0.0'
 default['consul']['config']['ports'] = {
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: attributes/default.rb [hunk 1]
```diff
@@ -4,19 +4,38 @@
...
 default['consul']['service_group'] = 'consul'
 
 default['consul']['bag_name'] = 'secrets'
 default['consul']['bag_item'] = 'consul'
 
-default['consul']['config']['path'] = '/etc/consul.json'
-default['consul']['config']['data_dir'] = '/var/lib/consul'
-default['consul']['config']['ca_file'] = '/etc/consul/ssl/CA/ca.crt'
-default['consul']['config']['cert_file'] = '/etc/consul/ssl/certs/consul.crt'
+case node['os']
```
johnbellone: This can be simplified quite a bit. There are two methods I am thinking of, the first being simply creating a temporary value for prefix directory and passing that through to all of the attributes. The second would be a heavier modification that would use interpolation later in the code. The [File class]() provides a platform-independent method for joining paths. I'll illustrate the latter (the former will require much more testing):
```ruby
prefix_path = case node['os']
              when 'linux', 'freebsd', 'mac_os_x'
                '/etc/consul'
              when 'windows'
                program_files
              end
default['consul']['config']['path'] = File.join(prefix_path, 'consul.json')
default['consul']['config']['data_dir'] = File.join(prefix_path, 'data')
default['consul']['config']['ca_file'] = File.join(prefix_path, 'CA', 'ca.crt')
```
The more complex solution would be lazily performing string interpolation. A helper method would need to do this so that we're able to correctly modify the path separators. I think the first approach is the most sensible right now as I don't immediately have a thought on how to do it elegantly.

libraries/consul_config.rb
```diff
@@ -114,49 +116,61 @@ def tls?
             [new_resource.ca_file, new_resource.cert_file, new_resource.key_file].each do |filename|
               directory ::File.dirname(filename) do
                 recursive true
-                owner new_resource.owner
-                group new_resource.group
-                mode '0755'
+                if node['os'].eql? 'linux'
```
johnbellone: I haven't used Chef with Windows yet. Is the **mode** the problem here?

libraries/helpers.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    def default_environment
+      {
+        'GOMAXPROCS' => [node['cpu']['total'], 2].max.to_s,
+        'PATH' => '/usr/local/bin:/usr/bin:/bin'
+      }
+    end
+
+    def command(config_file, config_dir)
```
johnbellone: Is this needed? I imagine we can do something in the separate provider classes themselves?

recipes/default.rb
```diff
@@ -5,38 +5,50 @@
 # Copyright 2014, 2015 Bloomberg Finance L.P.
 #
 
+node.default['nssm']['install_location'] = '%WINDIR%'
```
johnbellone: What's nssm?

libraries/consul_config.rb
Ginja: Only if you're not specifying an owner and/or group (which is what we're doing). Chef produces a warning.

libraries/helpers.rb
Ginja: I created this file to try and keep everything DRY. Some methods are only used once (e.g. default_environment) but others are used in several places (e.g. windows?, program_files). I prefer to keep provider & resource classes as clean as possible.

recipes/default.rb
Ginja: NSSM is [the Non-Sucking Service Manager](https://nssm.cc/). It's also what chocolatey used to run Consul on Windows. I removed chocolatey as we can't rely on the maintainers to keep Consul up-to-date. I also found using [chef-nssm](https://github.com/dhoer/chef-nssm) was a lot cleaner.

attributes/default.rb
Ginja: This is a good idea. I'll update my PR.

Ginja:
Rebased this PR to resolve conflicts.

gdavison:
@Ginja, this looks great. One problem I've found is that building all the file paths with `prefix_path` puts some things in the wrong place on Linux. For example, the Consul data should be under the `/var/lib/consul` hierarchy, rather than `/etc/consul`

Ginja:
@gdavison you're right, and it was briefly discussed changing that [here](https://github.com/johnbellone/consul-cookbook/pull/254#issuecomment-164091733). 

In order to keep everything clean & standardized on both platforms (i.e. consul data lives within the `data` folder underneath the prefix folder), I changed the values. I'd like to see it stay, but I can simply enforce that for my environment in our wrapper cookbook.

kamaradclimber:
@johnbellone  how can we help this PR getting merged?

gdavison:
+1 on merging. It works great on our Windows boxes

johnbellone:
Would you mind re-basing again?

attributes/default.rb
```diff
@@ -4,18 +4,23 @@
 #
 # Copyright 2014-2016, Bloomberg Finance L.P.
 #
+::Chef::Node.send(:include, ConsulCookbook::Helpers)
```
johnbellone: What's the purpose of this? The helpers should be injected automatically, no?

Ginja: They're not, unfortunately. I needed this to utilize the helper methods. I'll be rebasing shortly.

gdavison:
One issue: `consul_definition` resources have to have their path set manually on Windows. (And the Linux version doesn't use the `config_dir` parameter, but that's a separate issue)

Ginja:
I've got a PR incoming that'll fix that. Thanks for the head's up. I just noticed the nssm resource doesn't start services if they're stopped. So I added that check into the Windows provider. We should look at contributing some of these changes back to the [nssm cookbook](https://github.com/dhoer/chef-nssm).

legal90:
Be aware - this PR introduces a backward incompatibility:

```
# Previous default value
default['consul']['config']['data_dir'] = '/var/lib/consul'

# New default value
default['consul']['config']['data_dir'] = join_path data_prefix_path, 'data'
# e.q. '/var/lib/consul/data'
```

cc: @johnbellone

PR #263, opened 2016-01-21
Added ability to remove nssm parameters

### LINKED ISSUE #262: Idempotency
I'm not sure if anyone else is noticing this, but I'm seeing this cookbook consistently apply at least 3 resources (on both Windows & Linux hosts):

image

When I view what resources were changed in our monitoring tool, it looks like this:

image

I'm wondering if this is because of Poise. I haven't done a deep-dive into this issue as I wanted to bring it up here in case this is to be expected.

Ginja:
* You can now remove previously set nssm parameters by setting the hash values to `''` for Strings, & `0` for Integers. Also improved how the Consul service is stopped on Windows. If you don't stop it before you remove it, you end up in a SERVICE_PAUSED state.
* Fixes #262
* Fixed an edge case when stopping the service on Windows
* Removed problematic reload line. `new_resource.notifies(:reload, new_resource, :delayed)` causes the `consul_service` resource to reload every time Chef ran. This is an issue with the new sysvinit service provider when the service is in a stopped state when Chef runs. It would start and recieve a reload HUP too quickly & subsequently crash the service.
* Added attribute for specifying install_path for `consul_service` resource. My reasoning is that one should be able to change this crucial property without having to redeclare the whole resource.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -33,6 +33,8 @@ default['consul']['config']['ports'] = {
 default['consul']['diplomat_version'] = nil
 
 default['consul']['service']['config_dir'] = join_path config_prefix_path, 'conf.d'
+
+default['consul']['service']['install_path'] = windows? ? config_prefix_path : '/srv'
 default['consul']['service']['install_method'] = 'binary'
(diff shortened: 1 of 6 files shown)


johnbellone:
Would you mind rebasing?

Ginja:
Done.

Ginja:
The tests pass locally on my machine. It looks like a one-off issue.

PR #265, opened 2016-01-27
Create config files in config_dir

gdavison:
So that they can be found and loaded by the Consul agent

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,7 +32,7 @@ Metrics/AbcSize:
   Enabled: false
 PerceivedComplexity:
   Enabled: false
-SingleSpaceBeforeFirstArg:
+Style/SingleSpaceBeforeFirstArg:
   Enabled: false
 Style/ClassAndModuleChildren:
   Enabled: false
--- a/libraries/consul_watch.rb
+++ b/libraries/consul_watch.rb
@@ -11,12 +11,13 @@ module ConsulCookbook
     # @since 1.0.0
     class ConsulWatch < Chef::Resource
       include Poise(fused: true)
+      include ConsulCookbook::Helpers
       provides(:consul_watch)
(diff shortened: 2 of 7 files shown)


Ginja:
:+1: good catch.

libraries/consul_watch.rb
```diff
@@ -11,12 +11,13 @@ module Resource
...
     class ConsulWatch < Chef::Resource
       include Poise(fused: true)
+      include ConsulCookbook::Helpers
       provides(:consul_watch)
       default_action(:create)
 
       # @!attribute path
       # @return [String]
-      attribute(:path, kind_of: String, default: lazy { "/etc/consul/#{name}.json" })
+      attribute(:path, kind_of: String, default: lazy { join_path node['consul']['service']['config_dir'], "#{name}.json" })
```
johnbellone: I know that I am going to sound overly pedantic here, but I prefer to hardcode the defaults and keep the node object out completely. Is there any functional difference if you were to do that?

.rubocop.yml
```diff
@@ -32,7 +32,7 @@ Metrics/AbcSize:
   Enabled: false
 PerceivedComplexity:
   Enabled: false
-SingleSpaceBeforeFirstArg:
+Style/SingleSpaceBeforeFirstArg:
```
johnbellone: Great catch, I was scratching my head here when it spewed an error a few minutes ago but decided to leave it be.

johnbellone:
Take a look at my above comments. Let's get this merged and I think this is the last big obstacle for the 1.4 release.

gdavison:
I favoured the node value for a couple reasons:

1. We only have to set the default in one place
1. If someone wants to override the default, they don't have to add a `path` value to every consul resource they create

Not a huge issue to me either way, though.

Issue #235, opened 2015-11-13
does not start at boot on CentOS 6


petere:
On CentOS 6, this cookbook registers consul as an upstart job. The consul process is started when Chef runs. I can also manually start and stop consul using something like `sudo start consul`. But after rebooting, consul is not automatically started. I can start it manually again, and everything is fine, but that's probably not the idea.

The configuration in `/etc/init/consul.conf` seems to indicate that it should automatically start at system start.  But the upstart version CentOS 6 is so old, it's hard to debug anything.

Here is a small Vagrantfile to test this:

```ruby
Vagrant.configure('2') do |config|
  config.vm.box = 'chef/centos-6.6'
  config.vm.provision 'chef_solo' do |chef|
    chef.add_recipe 'consul-cluster'
    chef.json = {
      'consul' => {
        'config' => {
          'enable_syslog' => true
        }
      },
      'consul-cluster' => {
        'config' => {
          'bootstrap_expect' => 1
        }
      },
    }
  end
end
```

After `vagrant up`, check that consul is running, then run `vagrant reload`, and it won't come back up.

johnbellone:
What's the minor version of CentOS 6 that you're using?

petere:
6.6, as in the Vagrantfile shown above.

johnbellone:
I was able to confirm that Consul isn't starting on reboot.

legal90:
@petere FYI: it could be worked around by switching service provider to `sysvinit` (instead of `upstart`):
```ruby
node.set['poise-service']['consul']['provider'] = 'sysvinit'
```
Or the same with `poise_service_options` resource:
```ruby
poise_service_options node['consul']['service_name'] do
  _options['provider'] = 'sysvinit'
end
```

johnbellone:
That was what I was going to suggest as well. I'm not comfortable using the version of upstart on CentOS 6 because it is ancient. I'll try and take some time to mess around with this, but this is probably the best route.

petere:
I had tried that as well, but it doesn't work. The first problem is that the generated init script does not pass the path (`/usr/local/bin`) for consul, so it doesn't start. And if you fix that (manually), consul starts, but the init script fails because it checks for the PID of `/usr/bin/env`. And you don't get any error messages about either of these. It become clear very quickly that that scenario is not tested very well in poise-service. You can easily reproduce this by adding the appropriate attributes to the default test in `.kitchen.yml`.

johnbellone:
[...]

PR #266, opened 2016-01-28
Add custom template for "sysvinit" service provider

legal90:
The default template looks very tricky and it doesn't work well on RHEL 6: there are some artifacts in the console output while we are doing `service consul <stop | start>`:
https://github.com/poise/poise-service/blob/d9a5135dbe5711214c377c8ee2e7af029b1f10b7/chef/templates/default/sysvinit.sh.erb

This PR adds a custom template for `sysvinit` provider which looks pretty clear and works fine.
It was inspired by the init script form "consul-rpm" project: [consul.init](https://github.com/tomhillable/consul-rpm/blob/4f93c54a09067765418bd8c05926e816e610c1e8/SOURCES/consul.init)

Verified on CentOS 6.7 and Debian 7.9

P.s. This PR could be related to https://github.com/johnbellone/consul-cookbook/issues/235, but it doesn't fix this issue. For  RHEL/CentOS 6 we still have to set provider to `:sysvinit` in explicit to avoid the usage of `:upstart` (https://github.com/johnbellone/consul-cookbook/issues/235#issuecomment-172462543).

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -48,7 +48,7 @@ module ConsulCookbook
       if windows?
         %{agent -config-file="""#{config_file}""" -config-dir="""#{config_dir}"""}
       else
-        "/usr/bin/env consul agent -config-file=#{config_file} -config-dir=#{config_dir}"
+        "consul agent -config-file=#{config_file} -config-dir=#{config_dir}"
       end
     end
 
--- a/libraries/consul_service.rb
+++ b/libraries/consul_service.rb
@@ -161,6 +161,7 @@ module ConsulCookbook
         service.environment(new_resource.environment)
         service.restart_on_update(true)
         service.options(:systemd, template: 'consul:systemd.service.erb')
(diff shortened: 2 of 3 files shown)


johnbellone:
:+1:

libraries/helpers.rb
```diff
@@ -48,7 +48,7 @@ def command(config_file, config_dir)
       if windows?
         %{agent -config-file="""#{config_file}""" -config-dir="""#{config_dir}"""}
       else
-        "/usr/bin/env consul agent -config-file=#{config_file} -config-dir=#{config_dir}"
```
legal90: I've done this because `poise-service` splits the "command" on two parts by the first space:
https://github.com/poise/poise-service/blob/3355a9d1ae72761c3da36e72ed8b9912b8fb5be9/lib/poise_service/service_providers/sysvinit.rb#L53-L69

Thus, in the template we will be able to use variables `@daemon` and `@daemon_options`. The value of `@daemon` is something like `/usr/local/bin/consul`, e.q. it is already expanded to absolute path.

PR #269, opened 2016-01-29
Remove duplicated "data_dir" attribute

legal90:
I guess it was just a typo in #259
We already have `['config']['data_dir']`, so `['service']['data_dir']` is not needed. 

Otherwise it could cause a Consul agent start failure if you override only the one of these attributes.

cc: @Ginja

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -33,8 +33,6 @@ default['consul']['config']['ports'] = {
 default['consul']['diplomat_version'] = nil
 
 default['consul']['service']['config_dir'] = join_path config_prefix_path, 'conf.d'
-default['consul']['service']['data_dir'] = join_path data_prefix_path, 'data'
-
 default['consul']['service']['install_method'] = 'binary'
 default['consul']['service']['binary_url'] = "https://releases.hashicorp.com/consul/%{version}/%{filename}.zip" # rubocop:disable Style/StringLiterals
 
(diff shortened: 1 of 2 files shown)


Ginja:
Good catch! I override both of these values in our own environment. I'll change this in #263 in case @johnbellone wants to keep `['consul']['service']['data_dir']`.

legal90:
@Ginja Could you please explain what is the original purpose for `['consul']['service']['data_dir']` ?
Why don't you use ['consul']['config']['data_dir'] for that?

Ginja:
I can't unfortunately, I was simply modifying what was already there. I don't see a need for it, and am happy to eject it in my PR as well if @johnbellone would like.

legal90:
Well... I just thought that it was your idea, because this attribute was added in your commit: 
https://github.com/visioncritical/consul-cookbook/commit/ad2fcf5f291f0d325dc3f11d8530c7871dc8337a#diff-25e5d4a4446ae12a0d6f1162b6160375R36

Ginja:
Ha, wow. I've been working on this too long I don't even remember what I've done.

I added this so I could modify this attribute without overriding the whole resource. Makes it easier to use cross-platform:

https://github.com/visioncritical/consul-cookbook/commit/ad2fcf5f291f0d325dc3f11d8530c7871dc8337a#diff-e7d95569a41fab7bae4ab723c764c914L60
https://github.com/johnbellone/consul-cookbook/blob/master/recipes/default.rb#L55

scalp42:
Seeing the same issue here. Changed our data dir location.

Ginja:
I've updated #263 to change it back to just `data_prefix_path`, which is the original value.

PR #272, opened 2016-02-05
Correct acl creation

kamaradclimber:
Otherwise the diplomat gem raise an exception

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,7 +32,7 @@ Metrics/AbcSize:
   Enabled: false
 PerceivedComplexity:
   Enabled: false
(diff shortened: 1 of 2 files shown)


johnbellone:
Unfortunately our coverage here isn't the best. Would you be able to explain the change?

kamaradclimber:
`Diplomat::Acl.info` expect a third parameter which gives the behavior if the acl does not exist. The default behavior is to raise an exception.

See https://github.com/WeAreFarmGeek/diplomat/blob/master/lib/diplomat/acl.rb#L15, I'll put this into the commit message.

johnbellone:
Thanks!

PR #275, opened 2016-02-06
Windows data path

### LINKED ISSUE #273: Windows Consul service does not start up
The consul service expects its data directory to exist. This was recently removed, in commit aac210a, causing consul to fail on startup

gdavison:
Fixes #273 by creating Consul data directory

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -62,3 +62,20 @@ suites:
           acl_master_token: doublesecret
           acl_datacenter: fortmeade
           acl_default_policy: deny
+    excludes:
+      - windows-2012r2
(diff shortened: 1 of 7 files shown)


johnbellone:
Did #274 solve this?

gdavison:
It's fixed.

One style comment: `data_path_prefix` is not used as a prefix, so it should be renamed to `data_path`, and the Windows-specific appending of the `data` path segment should be moved from the attribute file to the `helpers.rb` file. (Or just merge in this PR :) )

Also, I have a bunch of integration tests in this PR that would be helpful

PR #279, opened 2016-02-16
Cope with diplomat behavior if acl does not exist

kamaradclimber:
Diplomat gem currently return an empty array if the acl does not exist.
This patch will allow to handle this behavior.


kamaradclimber:
duplicate of #278

PR #282, opened 2016-02-24
Replace nssm restart by powershell commands

kamaradclimber:
Very frequenty nssm commands (stop and start) fail because the service
gets into STOP_PENDING or START_PENDING state.
powershell commandlets such as restart-service have the good taste to
wait for the service to stop/start completely.

### DIFF
--- a/libraries/consul_service_windows.rb
+++ b/libraries/consul_service_windows.rb
@@ -63,20 +63,19 @@ module ConsulCookbook
                 action = v.eql?('') ? "reset consul #{k}" : "set consul #{k} #{v}"
                 batch "Set nssm parameter - #{k}" do
                   code "#{nssm_exe} #{action}"
-                  notifies :run, 'batch[Trigger consul restart]', :delayed
+                  notifies :run, 'powershell_script[Trigger consul restart]', :delayed
                 end
               end
-              batch 'Trigger consul restart' do
+              powershell_script 'Trigger consul restart' do
                 action :nothing
-                code "#{nssm_exe} restart consul"
+                code "restart-service consul"
               end
             end
             # Check if the service is running, but don't bother if we're already
             # changing some nssm parameters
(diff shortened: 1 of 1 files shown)


johnbellone:
I don't know too much about nssm to be honest. I did a little searching on the page and it seems like there are configurations where you can tweak some changes. 

http://nssm.cc/usage

Ginja:
I haven't seen this too much in our environment. Have you tried playing with the restart delay option of NSSM, @kamaradclimber?

> As of version 2.22, nssm can apply a mandatory delay between restarts of the application. This could be used, for example, to run a command at regular intervals such as an hourly batch script.

> To specify a restart delay, create an integer (REG_DWORD) value HKLM\System\CurrentControlSet\Services\servicename\Parameters\AppDelay and set it to the number of milliseconds to wait between restarts.

> The service will report its state as Paused while waiting for the next restart. Sending it a Continue control will temporarily cancel the delay and trigger a restart immediately.

> Please see the section on restart throttling below for notes on how nssm works when both throttling and restart delays are configured.

kamaradclimber:
thanks @Ginja for you reply. As far as I understand the restart delay option it only deals with a delay between the service stop and the moment nssm tries to start it again.

In my case, a simple restart command send to nssm is causing the state.

It seems to be due to consul taking some time to stop (writing some information to disk) that might be long (a few seconds for a 2k node cluster) and nssm does not wait properly for process termination thus declaring the service as "stop pending".

Apparently powershell commandlets handle this correctly.

Ginja:
@kamaradclimber oh, thanks for the clarification. That makes sense, I don't manage that large of a fleet in one Consul data centre, which probably would explain why I never ran into this issue. I just tested using those cmdlets locally & it looks like it'll work fine with NSSM.

I'd say go ahead & merge this, @johnbellone.

kamaradclimber:
thanks!

PR #289, opened 2016-03-16
add dev_mode config option

spheromak:
This option was added to consul 6.

---
This change is [Reviewable](https://reviewable.io/reviews/johnbellone/consul-cookbook/289)


spheromak:
@johnbellone as I am using the current branch in some recent work. I made this PR against install-resource. 

Also I am not sure if we want to add a constraint on the consul version being used to ensure this doesn't leak into an older install. I didn't see another example of this in the current library, and wasn't sure if it is something you consider in scope.

johnbellone:
@spheromak I don't think we yet have a clean way to limit configuration options for a specific version. That is something that I'll need to think a little more about. I have a general idea of how I want to approach refactoring the config resource, but need to finish up on the install one first.

johnbellone:
Would it be possible for you to rebase? I ended up doing so before merging that branch into master last night.

spheromak:
@johnbellone  you want me to reopen after rebase ?

johnbellone:
I added the attribute in the 2.1 release :).

PR #291, opened 2016-03-17
Adds custom resource for installing Consul.

### LINKED ISSUE #247: Refactor the population of TLS files to wrapper cookbooks?
Hi,

I'm noticing that consul_config.rb is currently expecting the TLS files to be set in a chef-vault. I fully support the use of chef-vault, and in fact I'm already using it; however, I'm adding the files to the filesystem in my wrapper cookbook, and I'd like to simply tell consul-cookbook where to look for the files once they're already there.

I much prefer this technique because it allows a lot of flexibility for myself and other consumers of this cookbook to handle secrets in their preferred way.

I'd be willing to write the pull request to make this happen, but I just wanted to make sure that this work would be appreciated before embarking on what I consider a fairly significant API change.

Let me know,

-Justin

### LINKED ISSUE #251: Updating consul version does not restart consul
Not sure if this is a bug for poise, libartifact, or the use of them both here. What I'm seeing is: when running the cookbooks and doing an upgrade from 0.5.2 -> 0.6.0 (binary install), the binary gets updated but the service isn't restarted. It looks like there's a subscription on config file changes, but there's no config file change in this process!

I looked to see if there was a convenient way to subscribe/notify on a binary update, but I'm not very familiar with poise or libartifact, and the fact that it's updating the binary inside an LWRP complicates things a bit.

johnbellone:
/cc @acaiafa @sh9189 
Closes #251 
Closes #247


legal90:
@johnbellone Thank you very much for this PR. It is a really big deal! :+1: 

But I've noticed that you've removed symlink creation (`/usr/local/bin/consul`). I just wondering - was there a significant reason for this?

Ginja:
@legal90 I believe it may be because it's tricky to orchestrate a symlink across providers, when not all platforms would use/need it (looking at you Windows). My guess is that @johnbellone decided to forgo the management of it as it's only referenced in the service script, and that the path to it will be updated automatically.

Ginja:
That being said, we should probably remove this default value:

https://github.com/johnbellone/consul-cookbook/blob/master/libraries/consul_service.rb#L48

PR #296, opened 2016-03-18
Moved user creation into separate recipe.

maf23:
This makes it much eaier to write wrapper cookbooks which need to create
config files owned by teh correct user. By having the users in a separate
recipe they can iclude that and get the user created before they create
the files. But still get the files created before consul is started.

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -33,17 +33,7 @@ if node['firewall']['allow_consul']
   end
 end
 
-unless windows?
-  group node['consul']['service_group'] do
-    system true
-  end
-
-  user node['consul']['service_user'] do
-    shell '/bin/bash'
-    group node['consul']['service_group']
-    system true
-  end
-end
+include_recipe 'consul::user'
 
 service_name = node['consul']['service_name']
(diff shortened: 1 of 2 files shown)


johnbellone:
What configuration files are you creating that can't get added after the whole recipe converges?

maf23:
A different data directory and certificates.

On Fri, Mar 18, 2016 at 8:45 PM, John Bellone <notifications@github.com>
wrote:

> What configuration files are you creating that can't get added after the
> whole recipe converges?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/johnbellone/consul-cookbook/pull/296#issuecomment-198515089>
>



-- 

Martin Forssén

Operations

Recorded Future <https://www.recordedfuture.com/>

+46 760 252357

maf@recordedfuture.com

johnbellone:
Why not just override the attributes to change both of those?

johnbellone:
```ruby
node.default['consul']['config']['data_dir'] = '/foo'
```

Issue #307, opened 2016-04-06
Consul service should restart instead of reload


miroswan:
If I make a change to a config, it doesn't show in the API or UI until I actually restart the service. Reload doesn't seem to suffice.

wstaples:
Hi @miroswan,

Can you tell us what configuration change you are making that doesn't seem to reload? 
I assume you are reloading the service and not using the CLI `consul reload`
According to the [docs](https://www.consul.io/docs/agent/options.html#reloadable-configuration) These are the only re-loadable items

### Reloadable Configuration

Reloading configuration does not reload all configuration items. The items which are reloaded include:

-Log level
-Checks
-Services
-Watches
-HTTP Client Address
-Atlas Token
-Atlas Infrastructure
-Atlas Endpoint

legal90:
@miroswan Automatic consul daemon restart could be really harmful. Please refer to the discussion https://github.com/johnbellone/consul-cookbook/issues/288

Issue #288, opened 2016-03-10
restart_on_update considered harmful


nicwaller:
This cookbook [restarts Consul](https://github.com/johnbellone/consul-cookbook/blob/f55f08bd3610d3b78e53ac96a3064adfe4192ca9/libraries/consul_service.rb#L160) if the configuration changes. That seems reasonable at first.

But Consul relies heavily on maintaining quorum; if quorum is not maintained, the Consul cluster goes into an outage. All that's needed is for `ceil(N/2)` servers to restart at the same time. Allowing Chef to restart Consul **servers** dramatically increases the likelihood of encountering this scenario. I believe that I have experienced this several times now. Here's what happens.

1. A new version of Consul is released. Nice!
1. Tweak attributes in Chef to obtain the latest version of Consul.
1. Send signal USR1 to chef-client on all nodes to immediately obtain the update.
1. Two or more Consul servers restart at the same time, and quorum is lost.

At that point, my only option is to rebuild the Consul cluster. Not good.

--

I recommend that this cookbook make one of the following changes:

- Never restart Consul **servers**. Let admins do that using other tools. Restarting agents is always safe.
- Add an attribute to toggle whether servers are restarted when configuration changes. This should default to `false` and come with a stern warning.
- Support some kind of cluster mutex (using Cluster, Redis, etc.) to guarantee safety when restarting servers

Isn't it strange that Consul doesn't have a built-in way of safely restarting individual servers, or even the entire cluster?

shortdudey123:
> Support some kind of cluster mutex (using Cluster, Redis, etc.) to guarantee safety when restarting servers

I have not used it before, but this is one thought ~ https://github.com/websterclay/chef-dominodes
`A Chef resource for mutual exclusion of blocks of recipe code. Useful for cross-cluster rolling restarts.`

johnbellone:
I am comfortable with making the change to using SIGHUP.

johnbellone:
I think that having any kind of cluster mutex is a code smell. I would much rather leave it to the system administrators to upgrade the client. What I'll probably end up doing is making sure that the configuration resource sends a reload (SIGHUP) and that updates to the installation don't actually change the running service.

jasonmcintosh:
Please note consul itself is designed to handle and support these use cases.  See examples below for a method which would prevent multiple nodes from restarting simultaneously.  
https://www.consul.io/docs/guides/semaphore.html
[...]

PR #309, opened 2016-04-18
Prevent "consul" service to be restarted on update

legal90:
https://github.com/johnbellone/consul-cookbook/issues/288 was not completely fixed by https://github.com/johnbellone/consul-cookbook/commit/80f4a1ce4edb1638482e2391dabd1ba547ed7fad because poise service option `restart_on_update` is still set to `true`.
It means that consul version update, causing init script re-rendering, will also cause consul service to be restarted. 
This PR prevents this behaviour.

Closes https://github.com/johnbellone/consul-cookbook/issues/307 (as "won't fix")

cc: @johnbellone

### DIFF
--- a/libraries/consul_service.rb
+++ b/libraries/consul_service.rb
@@ -87,7 +87,7 @@ module ConsulCookbook
         service.directory(new_resource.data_dir)
         service.user(new_resource.user)
         service.environment(new_resource.environment)
-        service.restart_on_update(true)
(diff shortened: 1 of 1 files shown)


Ginja:
Good catch, but I recommend waiting until johnbellone/vault-cookbook#52 is merged to see what comes out of the suggestions. Thanks, @legal90!

johnbellone:
:+1:

PR #316, opened 2016-05-20
Allow to configure consul_service behavior through attributes

### LINKED ISSUE #314: Question: How can a wrapper cookbook specify additional systemd options
On some setups, the LimitNOFILE option is required in the consul service definition.

Is it possible to add this option using the wrapper cookbook pattern?

So far, I've tried to specify attributes such as 
`default['consul']['service']['options']['template'] = 'consul-wrapper:mytemplate.erb'`
(and similar variation) without any luck.

kamaradclimber:
Fix #314

### DIFF
--- a/libraries/consul_service.rb
+++ b/libraries/consul_service.rb
@@ -46,6 +46,10 @@ module ConsulCookbook
       # The location of the Consul executable.
       # @return [String]
       attribute(:program, kind_of: String, default: '/usr/local/bin/consul')
+      # @!attribute restart_on_update
+      # Whether to restart the service on updates
+      # @return [Boolean]
+      attribute(:restart_on_update, kind_of: [TrueClass, FalseClass], default: false)
       # @!attribute options
       # Options passed to the underlying poise service.
       # @return [Hash]
@@ -89,7 +93,7 @@ module ConsulCookbook
         service.directory(new_resource.data_dir)
         service.user(new_resource.user)
         service.environment(new_resource.environment)
-        service.restart_on_update(false)
+        service.restart_on_update(new_resource.restart_on_update)
         service.options(:systemd, template: 'consul:systemd.service.erb')
         service.options(:sysvinit, template: 'consul:sysvinit.service.erb')
         service.options(:upstart, template: 'consul:upstart.service.erb', executable: new_resource.program)
--- a/libraries/consul_service.rb
+++ b/libraries/consul_service.rb
@@ -46,6 +46,10 @@ module ConsulCookbook
       # The location of the Consul executable.
       # @return [String]
       attribute(:program, kind_of: String, default: '/usr/local/bin/consul')
+      # @!attribute options
+      # Options passed to the underlying poise service.
+      # @return [Hash]
+      attribute(:options, kind_of: Hash, default: lazy { node['consul']['service']['options'] })
 
       def command
         "#{program} agent -config-file=#{config_file} -config-dir=#{config_dir}"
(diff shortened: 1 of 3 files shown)


johnbellone:
I want to rope in @coderanger to this discussion (as well as #314).

We have a lot of cookbooks that follow the pattern of implementing [PoiseService::ServiceMixin](https://github.com/poise/poise-service#servicemixin) as a custom resource, what's the best way to override service options like @kamaradclimber given that we set some defaults in [#service_options](https://github.com/johnbellone/consul-cookbook/blob/master/libraries/consul_service.rb#L85-L97)?

johnbellone:
Just for my thoughts:

I've been actually leaning a way from having a separate resource for service instances and instead using [poise_service](https://github.com/poise/poise-service#poise_service) directly where possible. The only caveat here is Windows, and I think that we can create a poise-service provider for nssm to be done with it once and for all.

johnbellone:
After [reading the documentation](https://github.com/poise/poise-service#service-options) it seems that we should be able to do this from a wrapper cookbook/policy:
```ruby
default['poise-service']['consul']['options']['systemd']['template'] = 'consul-wrapper:systemd.erb'
```

kamaradclimber:
@johnbellone any news regarding this MR ? Should I rebase and fix the conflicts or do you have another solution ?

kamaradclimber:
thanks for your answer.
we've used this branch of the cookbook for some time in production, it works as expected.

kamaradclimber:
done

libraries/consul_service.rb
```diff
@@ -46,6 +46,14 @@ class ConsulService < Chef::Resource
       # The location of the Consul executable.
       # @return [String]
       attribute(:program, kind_of: String, default: '/usr/local/bin/consul')
+      # @!attribute restart_on_update
+      # Whether to restart the service on updates
+      # @return [Boolean]
+      attribute(:restart_on_update, kind_of: [TrueClass, FalseClass], default: lazy { node['consul']['service']['restart_on_update'] })
```
johnbellone: Can we default this to false? This will still work as expected (using node attribute) because of [how we wire attributes into the service resource](https://github.com/johnbellone/consul-cookbook/blob/master/recipes/default.rb#L70-L72).

kamaradclimber: ok

libraries/consul_service.rb
```diff
@@ -85,10 +93,13 @@ def service_options(service)
         service.directory(new_resource.data_dir)
         service.user(new_resource.user)
         service.environment(new_resource.environment)
-        service.restart_on_update(false)
+        service.restart_on_update(new_resource.restart_on_update)
         service.options(:systemd, template: 'consul:systemd.service.erb')
         service.options(:sysvinit, template: 'consul:sysvinit.service.erb')
         service.options(:upstart, template: 'consul:upstart.service.erb', executable: new_resource.program)
+        new_resource.options.each do |key, value|
```
kamaradclimber: This is necessary to be able to override the template for systemd unit file (for instance to limit the number of possible open files)

kamaradclimber:
@johnbellone would you have feedback regarding this PR?

Ginja:
We have custom init/systemd scripts for all our Hashicorp related cookbooks, & this is how we override them:

```ruby
# our_wrapper_cookbook/recipes/default.rb
node.override['poise-service']['consul']['template'] = 'our_wrapper_cookbook:file.erb'
```

Issue #322, opened 2016-06-02
Restrict config write access


kingpong:
The consul cookbook makes the Consul config writable by the same user that runs the service, which is unnecessary and could be an attack vector. Would you accept a pull request which adds a couple new attributes and uses them for the ownership of `/etc/consul`?

    default['consul']['config_owner'] = 'consul'  # allows overriding with 'root' in wrapper
    default['consul']['config_group'] = 'consul'

johnbellone:
Perhaps we simply remove the attributes passing through in the default recipe (on this resource) and rely on the defaults. If you want to change to root/root you can override the config hash?

kingpong:
Ah! That didn't even occur to me. I think that's a great solution.

kingpong:
I may have spoke too soon. The conf.d directory is actually created by `consul_service`, not `consul_config` which means it is owned by the service user. Does conf.d need to be writable by the service user? If not, it could be moved into `consul_config` and belong to `node['consul']['config']['owner']`.

johnbellone:
I think it makes sense to move that directory and set permissions accordingly.

kingpong:
Fixed in #323 and #324.

nidhishopback:
Hi,

I am facing an issue related to config file permissions. I have overridden the default user of config and service to 'root' in wrapper cookbook : 
```
override['consul']['config']['owner'] = 'root'
override['consul']['config']['group'] = 'root'
override['consul']['service_user'] = 'root'
override['consul']['service_group'] = 'root'
```

Still, on running chef-client, it looks for consul user to add a service config :(
```
* file[/etc/consul/conf.d/chef-client.json] action create
      * cannot determine user id for 'consul', does the user exist on this system?
      ================================================================================
      Error executing action `create` on resource 'file[/etc/consul/conf.d/chef-client.json]'
      ================================================================================

      Chef::Exceptions::UserIDNotFound
      --------------------------------
      cannot determine user id for 'consul', does the user exist on this system?
```

Can you please suggest what I am missing here to have my config files owned by root?

PR #323, opened 2016-06-04
Prevent consul from writing its own config

kingpong:
This fixes issue #322.

The consul cookbook makes the Consul config writable by the same user
that runs the service, which is unnecessary and could be an attack
vector. This commit allows the user to specify a different set of
permissions for the configuration directories (/etc/consul) than for the
runtime data directory (/var/lib/consul).

The config owner still defaults to 'consul' since changing it to root is
a breaking change.

### DIFF
--- a/libraries/consul_config.rb
+++ b/libraries/consul_config.rb
@@ -20,10 +20,10 @@ module ConsulCookbook
       attribute(:path, kind_of: String, name_attribute: true)
       # @!attribute owner
       # @return [String]
-      attribute(:owner, kind_of: String, default: lazy { node['consul']['config']['owner'] })
+      attribute(:owner, kind_of: String, default: 'consul')
       # @!attribute group
       # @return [String]
-      attribute(:group, kind_of: String, default: lazy { node['consul']['config']['group'] })
+      attribute(:group, kind_of: String, default: 'consul')
       # @!attribute config_dir
       # @return [String]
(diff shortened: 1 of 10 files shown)


libraries/consul_config.rb
```diff
@@ -20,10 +20,13 @@ class ConsulConfig < Chef::Resource
       attribute(:path, kind_of: String, name_attribute: true)
       # @!attribute owner
       # @return [String]
-      attribute(:owner, kind_of: String, default: 'consul')
+      attribute(:owner, kind_of: String, default: lazy { node['consul']['config']['owner'] })
```
johnbellone: Any reason why we can't leave this simply as it was and rely on the value being passed through [in the default recipe](https://github.com/johnbellone/consul-cookbook/blob/master/recipes/default.rb#L54)?

kingpong: tl;dr: none at all. I've pushed a new commit to change it back and test a different way.

It took me a few minutes to understand why I did it this way. It's because I had already started updating [the Rspec](https://github.com/kingpong/consul-cookbook/blob/a5e2516354a5dc735999e761399efcdc1f314187/test/spec/libraries/consul_config_spec.rb#L16) based on my original suggestion in #322. I changed part of the implementation but forgot to change the other part since it didn't break the tests.  Anyway, it's reverted now and I'm testing the changes by using the resource property instead of an attribute.

PR #325, opened 2016-06-04
Misc bugfixes

kingpong:
Two bugfixes:
1. This quells a warning about "Option environment must be a kind of [String]" that gets emitted during rspec runs. Chef 13 will not allow the type mismatch.
2. The consul_service resource was using a default value attribute that was not configured anywhere. This changes it to use the same default as consul_config.

### DIFF
--- a/templates/default/sysvinit.service.erb
+++ b/templates/default/sysvinit.service.erb
@@ -64,7 +64,7 @@ _start() {
     daemon \
          --pidfile=$pidfile \
          --user=$user \
-         " { $exec <%= @daemon_options %> > $logfile 2>&1 & } ; echo \$! >| $pidfile "
+         " { $exec <%= @daemon_options %> >> $logfile 2>&1 & } ; echo \$! >| $pidfile "
      RETVAL=$?
      echo
      [ $RETVAL -eq 0 ] && touch $lockfile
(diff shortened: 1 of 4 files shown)


kingpong:
Not sure what's up with the build failure. Looking at the project build history it seems pretty random.

templates/default/sysvinit.service.erb
```diff
@@ -64,7 +64,7 @@ _start() {
     daemon \
          --pidfile=$pidfile \
          --user=$user \
-         " { $exec <%= @daemon_options %> &>> $logfile & } ; echo \$! >| $pidfile "
+         " { $exec <%= @daemon_options %> > $logfile 2>&1 & } ; echo \$! >| $pidfile "
```
johnbellone: What is your reason for doing this? This means that the log file will be overwritten.

kingpong: Whoopsie.  The reason for changing the syntax at all was to fix a test kitchen failure on CentOS 5. See the accompanying [git comment](https://github.com/johnbellone/consul-cookbook/pull/325/commits/49dbde591594ae6c158cb871db862ff060c89476) for more context.

The `>` instead of `>>` was totally accidental. Will fix momentarily and push.

PR #324, opened 2016-06-04
More permissions

kingpong:
This includes the branch from #323 but goes a step further and uses the configured attributes as defaults for the `consul_definition` and `consul_watch` resources. I made this a separate pull request since it wasn't something we discussed in #322.

PR #326, opened 2016-06-04
Updated readme with security paragraph

kingpong:
(This depends on #324)

Sorry for the barrage of PRs, just trying to make it easier for you to pick changes a la carte.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,22 @@ consul_service 'consul' do
   config_file config.path
 end
 ```
+
+### Security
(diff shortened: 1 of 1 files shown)


Ginja:
AFAICT you'll need to override `node['consul']['service_user']` to change the owner of the file, not `node['consul']['config']['owner']`, and this will only work on Linux (might want to mention that too).

https://github.com/johnbellone/consul-cookbook/blob/master/libraries/consul_config.rb#L125

PR #327, opened 2016-06-04
Use nologin for service user

kingpong:
Since the service user shouldn't be logging in with an interactive shell, set the shell to /sbin/nologin.

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -39,7 +39,7 @@ unless windows?
   end
 
   user node['consul']['service_user'] do
-    shell '/bin/bash'
+    shell '/sbin/nologin'
     group node['consul']['service_group']
     system true
   end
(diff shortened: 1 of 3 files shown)


Ginja:
This should be guarded against if the service user is `root`. Users probably don't want to change the shell for their `root` user. Could you add a `not_if` guard for this?

kingpong:
@Ginja Sure thing. PR amended.

johnbellone:
Hm, I probably should have thought a little about this prior to merging, any implications on BSD/Solaris?

kingpong:
Not sure about Solaris. I was unable to get test kitchen to run successfully on FreeBSD (see [this gist](https://gist.github.com/kingpong/9915265fb535df7fdded29b432893605)) but I did verify that FreeBSD supports nologin:

```
$ /sbin/nologin
This account is currently not available.
```

PR #328, opened 2016-06-12
Use correct nologin for distro

kingpong:
This fixes the nologin shell on debian-based systems since they use /usr/sbin instead of /sbin for this. The existing code behaves as expected (meaning it doesn't allow login) but the passwd entry points at a nonexistent file.

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -39,7 +39,7 @@ unless windows?
   end
 
   user node['consul']['service_user'] do
-    shell '/sbin/nologin'
+    shell %w( /sbin/nologin /usr/sbin/nologin /bin/false ).select { |f| ::File.exist?(f) }.first
     group node['consul']['service_group']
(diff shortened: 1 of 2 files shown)


johnbellone:
I am beginning to wonder if this logic should be broken out into a separate resource. I know that @coderanger has the [poise_service_user](https://github.com/poise/poise-service#poise_service_user) resource, is that something that fits our needs here?

kingpong:
@johnbellone I think `poise_service_user` does fit the need. It sets the shell to `/bin/false` which has the desired effect. The only benefit to using `nologin` is that it can present the user with a helpful message explaining why a shell was not opened.

coderanger:
That is what it's there for :)

johnbellone:
I will modify master to use poise_service_user as we mentioned above. Thanks @kingpong !

PR #330, opened 2016-06-17
Support of consul config reload on windows platform

jirkafajfr:
Hi,

I am trying to:

1. add support for ```action_reload``` on windows (simply by calling ```consul.exe reload```)
1. fix ```action_restart``` on windows (by wrapping restart block in ```notifying_block``` so its content gets executed)

### DIFF
--- a/libraries/consul_service_windows.rb
+++ b/libraries/consul_service_windows.rb
@@ -64,12 +64,20 @@ module ConsulCookbook
       end
 
       def action_reload
-        Chef::Log.info 'The service provider for Consul on Windows does not support reload!'
+        notifying_block do
+          execute 'Reload consul' do
+            command 'consul.exe reload'
+            cwd ::File.dirname(new_resource.program)
+            action :run
+          end
+        end
       end
 
       def action_restart
-        powershell_script 'Restart consul' do
-          code 'restart-service consul'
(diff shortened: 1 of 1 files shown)


libraries/consul_service_windows.rb
```diff
@@ -64,12 +64,20 @@ def action_enable
       end
 
       def action_reload
-        Chef::Log.info 'The service provider for Consul on Windows does not support reload!'
+        notifying_block do
+          execute 'Reload consul' do
+            command 'consul.exe reload'
```
Ginja: Looks good, but I'm assuming you've tested this. Does it interfere with NSSM at all? Since consul is a child process of NSSM, I'd worry that NSSM would see the "`HUP`" & try launching it again (effectively being a restart).

jirkafajfr: Hi, 

No, it does not interface with NSSM in any way. I guess that you're referencing to [consul reload article](https://www.consul.io/docs/commands/reload.html) and use of SIGHUP for config reload. 

My interpretation is that if NSSM would catch this signal and actually restarted service, usual startup lines like ```==> Starting Consul agent...``` would be presented in the logs. That is not happening during my tests.

PR #331, opened 2016-06-17
missing do in while loop upstart.service.erb line 51

mattp262:
The new upstart template works great in my testing.  It solved an issue I was running into with consul not coming up at boot.  I noticed however that there was a "do" missing in the post-start while loop.


johnbellone:
Nice catch!

PR #332, opened 2016-06-29
Start consul service on runlevel 2 in sysvinit

ZbigniewZabost-zanox:
Current LSB header of sysvinit script defines consul service to start on runlevel 3.

Default runlevel in Debian is 2 - https://wiki.debian.org/RunLevel, so it mean that when Debian based host is restarted, then consul service is not started.

To fix this issue I have update LSB header in sysvinit script so now consul service will be launched on runlevel 2.

### DIFF
--- a/templates/default/sysvinit.service.erb
+++ b/templates/default/sysvinit.service.erb
@@ -9,8 +9,8 @@
 # Provides:       <%= @name %>
 # Required-Start: $local_fs $network
 # Required-Stop:  $local_fs $network
-# Default-Start: 3 4 5
-# Default-Stop:  0 1 2 6
+# Default-Start: 2 3 4 5
+# Default-Stop: 0 1 6
 # Short-Description: Manage the consul agent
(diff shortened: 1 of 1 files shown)


johnbellone:
The [LSB second runlevel](https://en.wikipedia.org/wiki/Runlevel) does not have network interfaces up, correct?

MichaelKueller:
Yes it does. In debian and ubuntu it is the standard runlevel and defined as the "Full multi user mode".
Look here: https://debian-administration.org/article/212/An_introduction_to_run-levels

johnbellone:
@ZbigniewZabost-zanox you are saying that debian based systems don't get to run level 3? 

I don't really have a problem changing this but I want to make sure that we're not going to prematurely start the service when networking is not available.

MichaelKueller:
Debian based systems by default don't make a difference between runlevels 2 to 5. Therefor a standard debian installation does not get to run level 3.

PR #337, opened 2016-07-19
Add travis and docker integration tests.

johnbellone:
/cc @acaiafa @Ginja 

This adds the kitchen-bokken plugin and a mess of changes to travis.yml. It ultimately allows us to run integration tests using docker.


johnbellone:
/cc @coderanger

It looks like that [poise-service](https://github.com/poise/poise-service) is explicitly [disabling running systemd within a container](https://github.com/poise/poise-service/blob/master/lib/poise_service/service_providers/systemd.rb#L31). For now we'll converge this on those systems using sysvinit.

claudio-viola:
hi @johnbellone, just interested to understand why you chose kitchen-dokken over kitchen-docker , what are the advantages / disadvantages?
Thanks :) 👍

coderanger:
@claudio-viola Dokken supports systemd out of the box, kitchen-docker requires additional configuration to make it work.

PR #348, opened 2016-09-19
WIP: Update cookbook to use poise-archive for unzipping.

johnbellone:


legal90:
As I see, unpacking by `poise_archive` works weird. It tells that everything is OK:

```
      * directory[/opt/consul/0.7.0] action create (up to date)
      - unpack archive https://releases.hashicorp.com/consul/0.7.0/consul_0.7.0_linux_amd64.zip to /opt/consul/0.7.0
    * link[/usr/local/bin/consul] action create
      - create symlink at /usr/local/bin/consul to /opt/consul/0.7.0/consul
```

But `/opt/consul/0.7.0/` is empty after that. Reproduced on the following suites defined in `.kitchen.dokken.yml`:
- default-centos-6
- default-centos-7

metadata.rb
```diff
@@ -23,8 +23,8 @@
 depends 'golang'
 depends 'firewall', '~> 2.0'
 depends 'poise', '~> 2.2'
+depends 'poise-archive', '~> 1.3'
 depends 'poise-service', '~> 1.4'
-depends 'rubyzip', '~> 1.3'
```
legal90: You've removed `rubyzip`, but `consul_installation_binary` resource is still using `zipfile` resource from it. So, it's broken and ChefSpec test fails.

Issue #271, opened 2016-02-02
consul service user /bin/false shell ?


epcim:
I have testes 1.4 version from master and found the service_user created "consul" is being OOTB created with /bin/false default shell. Which causes all scripted checks effectively silent.

`poise_service_user` doesn't have the "shell" attribute. So the only way to make the user with effective shell is to create him first. https://github.com/poise/poise-service#poise_service_user

As I am quite new with consul I am not about to post PR as this might be resolved multiple ways.

johnbellone:
Which checks are failing?

epcim:
I mean consul_definition, type=check. In other words any script execution: parameters(script:'xyz.sh'). As the "service consul user" don't have a proper shell configured.

I workaround it by: 
{{{
# change the user shell "so it may run scripts for checks"
user node['consul']['service_user'] do
  shell '/bin/bash'
  action :modify
end
}}}

johnbellone:
Ah, interesting. It makes sense that this doesn't work just wonder why we never caught it before.

johnbellone:
This was fixed in recent patch.

PR #349, opened 2016-09-20
Allow specification of shell for the consul user

jirkafajfr:
Hi,

following two commits are changing shell of the consul user to /bin/nologin (or its variations)
- [Use nologin for service user](https://github.com/johnbellone/consul-cookbook/commit/938357ac844e0b6df66d92142957cfa6a20393a0)
- [Modify user/group creation to use poise-service-user](https://github.com/johnbellone/consul-cookbook/commit/68bc28a57add628476e05851db6412f1cbd72361)

this configuration prevents consul from executing handler in [consul watch](https://www.consul.io/docs/agent/watches.html) with following error:

```
2016/09/20 21:33:02 [ERR] agent: Failed to invoke watch handler 'XYZ': exit status 1
```

This patch allows wrapper cookbook to specify shell explicitly while retaining "security" for people not using consul watch functionality.

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -33,13 +33,15 @@ if node['firewall']['allow_consul']
   end
 end
 
+service_name = node['consul']['service_name']
 poise_service_user node['consul']['service_user'] do
   group node['consul']['service_group']
+  shell node['consul']['service_shell'] unless node['consul']['service_shell'].nil?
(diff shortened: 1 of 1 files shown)


quulah:
There's also the old issue #271. So the scripts checks seem to be broken too without the user's shell.

Issue #351, opened 2016-10-03
There is no longer a symlink to consul in /usr/local/bin


willejs:
Since 2.0 there is no longer a symlink of consul bin in the path.

PR #352, opened 2016-10-03
symlink /usr/local/bin/consul to the install location

willejs:
This fixes issue https://github.com/johnbellone/consul-cookbook/issues/351


kamaradclimber:
This patch breaks windows installation

PR #353, opened 2016-10-04
Fix installation on arm machines

rollbrettler:
There is no `armv7l` package only a `arm` package, therefore this throws an error.

### DIFF
--- a/libraries/consul_installation_binary.rb
+++ b/libraries/consul_installation_binary.rb
@@ -86,6 +86,7 @@ module ConsulCookbook
         case node['kernel']['machine']
         when 'x86_64', 'amd64' then ['consul', resource.version, node['os'], 'amd64'].join('_')
         when /i\d86/ then ['consul', resource.version, node['os'], '386'].join('_')
+        when /^arm/ then ['consul', resource.version, node['os'], 'arm'].join('_')
         else ['consul', resource.version, node['os'], node['kernel']['machine']].join('_')
         end.concat('.zip')
       end


libraries/consul_installation_binary.rb
```diff
@@ -86,6 +86,7 @@ def self.binary_basename(node, resource)
         case node['kernel']['machine']
         when 'x86_64', 'amd64' then ['consul', resource.version, node['os'], 'amd64'].join('_')
         when /i\d86/ then ['consul', resource.version, node['os'], '386'].join('_')
+        when /^arm/ then ['consul', resource.version, node['os'], 'arm'].join('_')
```
legal90: What's the exact value of `node['kernel']['machine']` in your case? Can we just match it with a string `arm`, without regex?

rollbrettler: Sorry but I think I don’t get your question? Maybe I should elaborate on what my problem was.

I was trying to use this cookbook to install consul on a raspberry pi. Doing so threw an error “403 Forbidden” since the package url was `…/consul_0.7.0_linux_armv7l.zip`. This is not the official downloadable package path. Its simply `…/consul_0.7.0_linux_arm.zip`.

Also it has to be a regex since there are different arm versions but they all start with `arm…`.

Does that answer your question? I am also happy to change if I got something wrong here.

legal90: > Also it has to be a regex since there are different arm versions but they all start with arm…

Ah, then it looks good to me 👍 
Yes, you've answered my question. Thank you!

rollbrettler:
@legal90 Whats with the failing travis checks. I saw that master is failing as well.

legal90:
I'm not sure, but https://github.com/johnbellone/consul-cookbook/pull/348 might be a reason of Travis CI build failures.

cc: @johnbellone

legal90:
@rollbrettler Thanks for the contribution!

PR #359, opened 2016-10-09
Added missing attributes

matelukas:


freimer:
I haven't had a chance to look deep into the build failures yet, but these appear cosmetic to the change.  zipfile missing, space at end of line, some package install failure on Debian.   Any chance of getting this fixed up and merged soon?  I'll need at least the performance attribute.

Any thought of turning this into a new-style resource, instead of a LWRP?  I can see the value of validating the config options, but this also requires this cookbook to closely track the consul product.  If it were a new-style resource then it could include a JSON schema and use that to check, and also have an option to turn validity checking off if the product includes some new feature.  That would allow the cookbook to still work without having to be immediately updated.  Just some thoughts...

libraries/consul_config.rb
```diff
@@ -79,6 +80,7 @@ class ConsulConfig < Chef::Resource
       attribute(:rejoin_after_leave, equal_to: [true, false], default: true)
       attribute(:server, equal_to: [true, false], default: true)
       attribute(:server_name, kind_of: String)
+      attribute(:session_ttl_min, kind_of: String)
```
legal90: Please, delete the trailing whitespace in this line

freimer: I got the diff of the commit and applied it to my repo, then fixed the whitespace and committed.  I can't get that to this PR, right?  @matelukas needs to do that, right?

legal90:
I've appended your brunch with 8e1396f, adding more attributes appeared in Consul v0.7.0.
@matelukas, thank you for the contribution!

freimer:
Do we need to bump the version in metadata.rb to 2.2.1?

legal90:
IMO, the version bump should happen simultaneously with a new version release, which is up to @johnbellone.

freimer:
OK, gotcha.  Thanks!

PR #362, opened 2016-10-26
Use config user/group for consul_definition instead of hard coded string

rollbrettler:
Does anything speak against using the `node['consul']['config']['group']` and `node['consul']['config']['user']` value for the default consul_definition attributes?

### DIFF
--- a/libraries/consul_config.rb
+++ b/libraries/consul_config.rb
@@ -20,10 +20,10 @@ module ConsulCookbook
       attribute(:path, kind_of: String, name_attribute: true)
       # @!attribute owner
       # @return [String]
-      attribute(:owner, kind_of: String, default: 'consul')
+      attribute(:owner, kind_of: String, default: lazy { node['consul']['service_user'] })
       # @!attribute group
       # @return [String]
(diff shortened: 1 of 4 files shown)


johnbellone:
Did you try setting `node['consul']['service_user']` and `node['consul']['service_group']`? I don't really see why you would need to do this with those node attributes.

rollbrettler:
I changed it to `node['consul']['service_user']` and `node['consul']['service_group'] ` but maybe I didn’t fully get your suggestion yet.

rollbrettler:
@johnbellone https://github.com/johnbellone/consul-cookbook/issues/372 is basically what I ran into as well and what I tried to fix with this PR

mishak87:
Watch and service resources are still using hardcoded strings `'consul'`.

https://github.com/johnbellone/consul-cookbook/blob/cd562a7/libraries/consul_watch.rb#L22-L28
https://github.com/johnbellone/consul-cookbook/blob/cd562a7/libraries/consul_service.rb#L24-L31

legal90:
@mishak87 Thanks! Pull-Request #445 fixes that

PR #364, opened 2016-11-03
Filter out bootstrap and bootstrap_expect parameters in client configs

gdavison:
Filter out some server-specific parameters from client config files. Consul 0.7 refuses to start when, for instance, `bootstrap_expect` is present.

When we set up our development environments, we set the parameter on the whole Chef environment, as the number of Consul servers seems to fit as a property of the environment. So when I tried updating one of the dev environments today, each client machine failed.

I haven't tested which other parameters will cause failure, if any


gdavison:
Blocked by failing RSpec tests on master

PR #368, opened 2016-11-07
Fixed service user specs

gdavison:
Added tests for service user creation to ensure that default user is not created and verify that nothing is created for Windows.

Sets individual web_ui specs to pending, so that they are ignored.

Closes johnbellone#365


legal90:
Thanks!

PR #371, opened 2016-11-09
[config] add dogstatsd_addr configuration attribute

theckman:
Consul can send metrics to the Datadog Agent using DogStatsD metrics.
This adds support for configuring Consul to use `dogstatsd` metrics with
the cookbook.


legal90:
@theckman Thank you for your contribution! But seems like we don't need to add this config option to `consul_config` resource (see the comment above). 
So, I close this PR.

PR #380, opened 2016-11-29
Add the following parameters in consul config supported in version 0.…

iliasbertsimas:
…7.1 and after:
- raft_multiplier
- reap
- retry_join_ec2 (Hash)
- serf_lan_bind
- serf_wan_bind


libraries/consul_config.rb
```diff
@@ -69,10 +69,13 @@ class ConsulConfig < Chef::Resource
       attribute(:key_file, kind_of: String)
       attribute(:leave_on_terminate, equal_to: [true, false], default: false)
       attribute(:log_level, equal_to: %w(INFO DEBUG WARN), default: 'INFO')
+      attribute(:max_stale, kind_of: String)
```
legal90: The documentation says that `max_stale` is a subset of `dns_config`, not a standalone option.
https://www.consul.io/docs/agent/options.html#max_stale

libraries/consul_config.rb
```diff
@@ -82,7 +85,10 @@ class ConsulConfig < Chef::Resource
       attribute(:retry_interval_wan, kind_of: String)
       attribute(:retry_join, kind_of: Array)
       attribute(:retry_join_wan, kind_of: Array)
+      attribute(:retry_join_ec2, kind_of: [Hash, Mash])
```
legal90: Following the alphabet order, then it should be defined _before_ `retry_join_wan` ;)

PR #385, opened 2016-12-15
Fit kitchen test failures

gdavison:
The `client` `serverspec` tests were failing because the `consul` version was not updated. Moved the version definition to `spec_helper.rb`.

Zip installation does not seem to be implemented anymore, so removed the tests


legal90:
@gdavison I've rebased your branch and fixed conflicts. Thanks again!

PR #395, opened 2017-01-12
Remove firewall dependency

### LINKED ISSUE #392: Can't use the cookbook with Chef 11 because of firewall cookbook
Hi,
  I am trying to use the cookbook Chef11 but got error with firewall cookbook because firewall cookbook 2.0.0 and above can only be used with Chef 12.  I wonder if consul 2.0 and above has to use firewall cookbook 2.0.0 or can use firewall cookbook with an earlier version?

legal90:
It was discussed in https://github.com/johnbellone/consul-cookbook/issues/392. According to the best practices for application cookbooks, it's better to manage firewall setting outside of this cookbook.

**NB!** This PR introduces breaking changes. It is recommended to merge it alongside with the major version bump.

Closes #392


legal90:
#397 should fix the TravisCI builds

PR #404, opened 2017-01-24
Implement home directory creation for consul user

sun77:
Poise-service doesn't create home directory with poise_service_user resource
Also allow us to change the home directory with an attribute

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -6,8 +6,19 @@
 #
 node.default['nssm']['install_location'] = '%WINDIR%'
 
+consul_home = node['consul']['service_user_home']
 service_name = node['consul']['service_name']
+
+# poise_service_user resource doesn't create home directory
+directory consul_home do
+  owner node['consul']['service_user']
+  group node['consul']['service_user']
+  mode '0755'
+  only_if { node['os'] == 'linux' }
+end
+
 poise_service_user node['consul']['service_user'] do
+  home node['consul']['service_user_home']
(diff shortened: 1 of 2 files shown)


recipes/default.rb
```diff
@@ -6,8 +6,19 @@
 #
 node.default['nssm']['install_location'] = '%WINDIR%'
 
+consul_home = node['consul']['service_user_home']
```
legal90: I see that we use this variable only once. May be, we can omit that and just use an attribute as is?

legal90:
@Ginja @johnbellone What do you think about this? Should we manage the home dir in this cookbook by default? 
I'm not sure about using `directory` resource for that. I would roll `poise_service_user` back to [`user` + `group` resources](https://github.com/johnbellone/consul-cookbook/blob/v2.1.3/recipes/default.rb#L36-L46) with setting `manage_home true`. Then the home dir will be created by "useradd" with system-defined attributes and permissions.

Ginja:
https://github.com/poise/poise-service/issues/9

Why does the Consul user need a home? I'm in agreeance with @coderanger on this one. Generally service users don't have homes for security reasons. If one is required, I would put that logic in a wrapper cookbook.

kamaradclimber:
consul user is the one used when launching health checks of type `script`. Many scripts/binary assume the existence of a $HOME

johnbellone:
@kamaradclimber What would be an example?

PR #408, opened 2017-02-02
Update Consul support to 0.7.4, some bug fixes

jasonmcintosh:
Added 0.7.3 binary sums
Changed default to 0.7.3 
Removed  freebsd binary sums
Fixed some source install  - go version 1.7 requirement, build path, zip dependency

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -32,7 +32,7 @@ default['consul']['diplomat_version'] = nil
 
 default['consul']['service']['config_dir'] = join_path config_prefix_path, 'conf.d'
 
-default['consul']['version'] = '0.7.2'
+default['consul']['version'] = '0.7.4'
 
 # Windows only
 default['consul']['service']['nssm_params'] = {
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -46,6 +46,8 @@ suites:
           bootstrap_expect: 1
   - name: git
     attributes:
+      go:
+        version: 1.7
       consul:
         config: *default-config
         installation:
--- a/README.md
+++ b/README.md
@@ -206,6 +206,11 @@ consul_execute 'uptime' do
   options(service: 'api')
 end
 ```
+### Warning on git based installs
+Consul states that Go 1.7 is a requirement. The default go installation uses
+1.5, so you may need to override a `['go']['version']` attribute to allow the
+git installation to work reliably.
+
 
 All of the [options available on the command-line][12] can be passed
 into the resource. This could potentially be a *very dangerous*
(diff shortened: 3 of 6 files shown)


metadata.rb
```diff
@@ -4,7 +4,7 @@
 license 'Apache 2.0'
 description 'Application cookbook which installs and configures Consul.'
 long_description 'Application cookbook which installs and configures Consul.'
-version '2.3.1'
+version '2.3.2'
```
legal90: Usually, it's not recommended to bump the cookbook version in the scope of the Pull-Request because it's up to the maintainer how & when it should be bumped and released.

attributes/default.rb
```diff
@@ -4,7 +4,7 @@
 #
 # Copyright 2014-2016, Bloomberg Finance L.P.
 #
-
+default['go']['version'] = '1.7'
```
legal90: Do we really need to hard-code this version here? 
It's highly undesirable for community cookbooks to override external attributes.

attributes/default.rb
```diff
@@ -32,7 +32,7 @@
 
 default['consul']['service']['config_dir'] = join_path config_prefix_path, 'conf.d'
 
-default['consul']['version'] = '0.7.2'
+default['consul']['version'] = '0.7.3'
```
legal90: May be we should set it to 0.7.4 now?

metadata.rb
jasonmcintosh: Okie dokie - happily will remove that change.

attributes/default.rb
jasonmcintosh: Another option would be to update the kitchen tests to specify go 1.7, and then have a check to make sure go 1.7 is installed?  It's listed as a flat our requirement on the consul website.  When I did a clean checkout, tests were completely failing to do any git based installs.  This was one of the things I'd seen to fix it (the other was that zip was missing).

attributes/default.rb
jasonmcintosh: Just pushed out the change for that :)

jasonmcintosh: Well, pushing out a change - was half way through.  Update should be here shortly on this.

attributes/default.rb
legal90: Pinging @johnbellone: Do you have any thoughts regarding this?

jasonmcintosh: I'm trying a few things, retesting - removing that, running tests, had git-centos-* failures similar:
```
    command/agent/command.go:39:2: cannot find package "github.com/hashicorp/go-multierror" in any of:
           	/usr/local/go/src/github.com/hashicorp/go-multierror (from $GOROOT)
           	/opt/go/src/github.com/hashicorp/go-multierror (from $GOPATH)
```
Testing a few other things right now.

jasonmcintosh:
K - I THINK that I've got all the changes we've talked about worked out.  This includes the move of the default attribute for go to the kitchen test, an update to the README to warning about the go dependency version issue for the git installation versions, as well as the metadata.rb reverted to the prior version.

legal90:
Thanks! I've just rebased the branch, so let's wait for a TravisCI build

PR #415, opened 2017-03-03
fixing permission of generated config

dennybaa:
There's no need in such strict options for a configuration file. However if there is it should be configurable for an enduser.

### DIFF
--- a/libraries/consul_definition.rb
+++ b/libraries/consul_definition.rb
@@ -60,7 +60,7 @@ module ConsulCookbook
             unless platform?('windows')
               owner new_resource.user
               group new_resource.group
-              mode '0640'
+              mode '0644'
             end
           end
         end
--- a/test/integration/default/default_spec.rb
+++ b/test/integration/default/default_spec.rb
@@ -73,7 +73,7 @@ describe file("#{confd_dir}/consul_definition_check.json") do
   it { should be_file }
   it { should be_owned_by 'root' }
   it { should be_grouped_into 'consul' }
-  its('mode') { should cmp '0640' }
+  its('mode') { should cmp '0644' }
 end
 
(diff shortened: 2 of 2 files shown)


legal90:
Hi @dennybaa,
Do you have any arguments why "there is no such need"? 
Configuration file might contain secrets like encryption key, master token or AWS credentials. 
The last two are not recommended to keep in the configuration file but anyway it should not be readable by other system users, IMO.

cc: @johnbellone

dennybaa:
@legal90 Hi, yes, please correct me if I'm wrong.

We are referring to a service or a check config not the main config, since the PR has nothing to do with it, right?
Consul doesn't require any parameters which should be sensitive: [services](https://www.consul.io/docs/agent/services.html) and [checks](https://www.consul.io/docs/agent/checks.html) definitions.
So in 80% of cases users won't put anything sensitive into these files, for other 20% the **mode** might be configurable.

As for me now I couldn't reuse this cookbook for services configs generation, because the mode is strict and hardcoded :( In my case consul user/group is not known or viable during configs generation.

PS. I'm using the cookbook **solely** for services/checks configs generation.

legal90:
Ah, now got it. 
Sorry, I've misunderstood it at first and thought that it was about the consul service configuration (`/etc/consul/consul.json`), not check's config.

Then it looks good to me. But please, fix the TravisCI tests - this line should be also changed to "644": https://github.com/johnbellone/consul-cookbook/blob/49b4b3f/test/integration/default/default_spec.rb#L76

dennybaa:
@legal90 well, np. I also suggest to update watch permissions :), though I'm not using em. `0640` also seems to much for a watch file.

legal90:
Thanks 👍

PR #419, opened 2017-03-21
added missing ACL config options

davidsayers:
added missing ACL config options


legal90:
@davidsayers Thank you! 👍

PR #421, opened 2017-03-23
Consul 0.7.5 & Windows Fixes

Ginja:
* Bump default version to 0.7.5
* Fixed upgrading issue on Windows
  * The NSSM service would not use the new binary location
  * Added extends to certain resources to ensure we're using the Consul Cookbook helper methods
* Added SHELL environment variable to default_environment hash
  * To make it easier to troubleshoot why script checks don't fire when consul is launched with systemd
* Re-enabled the Windows Service spec tests as they seem to work locally for me now

### DIFF
--- a/libraries/consul_service.rb
+++ b/libraries/consul_service.rb
@@ -55,6 +55,7 @@ module ConsulCookbook
         {
           'GOMAXPROCS' => [node['cpu']['total'], 2].max.to_s,
           'PATH' => '/usr/local/bin:/usr/bin:/bin',
+          'SHELL' => node['consul']['service_shell'],
         }
       end
     end
--- a/libraries/consul_service.rb
+++ b/libraries/consul_service.rb
@@ -51,12 +51,16 @@ module ConsulCookbook
         "#{program} agent -config-file=#{config_file} -config-dir=#{config_dir}"
       end
 
+      def shell_environment
+        shell = node['consul']['service_shell']
+        shell.nil? ? {} : { 'SHELL' => shell }
+      end
+
       def default_environment
         {
           'GOMAXPROCS' => [node['cpu']['total'], 2].max.to_s,
-          'PATH' => '/usr/local/bin:/usr/bin:/bin',
-          'SHELL' => node['consul']['service_shell'],
-        }
+          'PATH' => '/usr/local/bin:/usr/bin:/bin'
+        }.merge(shell_environment)
       end
     end
   end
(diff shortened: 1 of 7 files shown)


libraries/consul_service.rb
```diff
@@ -55,6 +55,7 @@ def default_environment
         {
           'GOMAXPROCS' => [node['cpu']['total'], 2].max.to_s,
           'PATH' => '/usr/local/bin:/usr/bin:/bin',
+          'SHELL' => node['consul']['service_shell'],
```
legal90: Will it work fine if `node['consul']['service_shell']` is not defined? (`nil`)

Ginja: Yup! It has a default value of /bin/nologin if present, or /bin/false

https://github.com/poise/poise-service/blob/master/README.md#attributes-2

Ginja: D'oh. Good point. I'll add a check here.

Ginja: @legal90, question for you.

At this point, do we want to make this separate value from `node['consul']['service_shell']`? If we do, the user could still have no valid shell and we could define a separate shell for when we're running in systemd. The value could either be hardcoded or another attribute.

legal90: Hm... I'd vote for `node['consul']['service_shell']` because the name of this argument os exactly what we expect here. But let's ask @johnbellone too. 

But in any case, the element should be appended using `if` condition, since the attribute is not defined by default. For example:
```
def default_environment
  env = {
    'GOMAXPROCS' => [node['cpu']['total'], 2].max.to_s,
    'PATH' => '/usr/local/bin:/usr/bin:/bin',
  }
  env['SHELL'] = node['consul']['service_shell'] if node['consul']['service_shell']
  env
end
```

Ginja: Unless @johnbellone disagrees, I'll just use `node['consul']['service_shell']` with a condition. If others, like us, want to specify a custom shell in the systemd file so the `consul` user can run script checks, they can override the template. I'll look at doing this later this week.

legal90: @Ginja Yeah, that looks good. Could you please update your PR by adding a condition?

legal90:
@Ginja Could you please also fix the style offense?
https://travis-ci.org/johnbellone/consul-cookbook/jobs/220580777#L504-L506

legal90:
@Ginja Thank you! I've rebased your branch in order to fix merge conflict.
TravisCI failure should be fixed by #424

Ginja:
Thanks, @legal90!

PR #422, opened 2017-04-04
Support nssm cookbook 3.0

Annih:
Nssm cookbook version 3.0 has been release with a breaking change.
The 'params' property has been renamed in 'parameters' for Chef 13 compat.

This code use 'parameters' if the resource expose this property and 'params'
otherwise.

**Cc:** @aboten


legal90:
@Annih Thank you for the contribution!

PR #424, opened 2017-04-07
attributes: explicit node.default server->false

### LINKED ISSUE #423: Default recipe starts consul with `server: true`
seemingly, the root cause is that consul version at least > 0.7.0 ships with `server: true` as a default in `consul.json`;

this can be seen by changing `node['consul']['version'] = 0.7.0` in the latest cookbook version, and noticing no `server: true` line in said file as compared to the cookbook's default consul version.

The docs are out of date WRT to this change in behaviour.

vsudilov:
closes #423 

Wasn't sure how you'd prefer to add tests, since server:false is explicitly set in `.kitchen.yml`. Happy to iterate on that as needed.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -37,13 +37,10 @@ suites:
           acl_default_policy: deny
     excludes:
       - windows-2012r2
+
   - name: client
     named_run_list: client
-    attributes:
-      consul:
-        config:
-          server: false
-          bootstrap_expect: 1
+
   - name: git
     attributes:
       go:
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -10,6 +10,7 @@ default['consul']['service_user'] = 'consul'
 default['consul']['service_group'] = 'consul'
 default['consul']['create_service_user'] = true
 
+default['consul']['config']['server'] = false
 default['consul']['config']['owner'] = 'consul'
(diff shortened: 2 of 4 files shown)


legal90:
@vsudilov Thanks for your contribution! 
However, since the default value is currently hard-coded in the resource, I think we should fix it there, rather then overriding via node's attribute:

https://github.com/johnbellone/consul-cookbook/blame/7b9b75c86ba4963dce6bebd94f289b258942062a/libraries/consul_config.rb#L92

vsudilov:
Yep, agree that's a much better solution -- commit added.

legal90:
NB: The TravisCI run [reveals](https://travis-ci.org/johnbellone/consul-cookbook/jobs/219702766) the very important issue which happens after the removal the default value of `"server"` attribute (resource parameter). 

Since we access this parameter [here](https://github.com/johnbellone/consul-cookbook/blob/master/libraries/consul_config.rb#L191), it becomes defined with the default value `nil`, so it is rendered to JSON config file as `"server": null`, which is not valid.
I've fixed that by removing `nil` values right before rendering the JSON config file. 
@vsudilov I've updated this PR by adding commits to your branch. Now Travis build is green

@johnbellone Please, review this PR.

vsudilov:
Thanks :)

PR #425, opened 2017-04-07
add support for specifying the raft protocol version

poblahblahblah:
consul 0.8.0 includes some autopilot features look really nice. In order to enable these you need to specify the raft protocol version.

This change allows us to do that.

### DIFF
--- a/libraries/consul_config.rb
+++ b/libraries/consul_config.rb
@@ -76,6 +76,7 @@ module ConsulCookbook
       attribute(:performance, kind_of: [Hash, Mash])
       attribute(:ports, kind_of: [Hash, Mash])
       attribute(:protocol, kind_of: String)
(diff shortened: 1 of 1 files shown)


libraries/consul_config.rb
```diff
@@ -76,6 +76,7 @@ class ConsulConfig < Chef::Resource
       attribute(:performance, kind_of: [Hash, Mash])
       attribute(:ports, kind_of: [Hash, Mash])
       attribute(:protocol, kind_of: String)
+      attribute(:raft_protocol, kind_of: Integer, default: false)
```
legal90: I think we should not specify `default: false` here because Consul can handle the default value by itself.

poblahblahblah:
@legal90 sounds good - PR updated to remove the default.

legal90:
@poblahblahblah Thank you!

PR #426, opened 2017-04-23
Update unit tests configuration

legal90:
Currently we have unit tests failed because of a `cheffish` gem conflict (is required by `chef-dk` gem). This PR removes some unnecessary gem dependencies and introduces Chef Delivery CLI for running tests, like Chef Software does that for their cookbooks.

* Use Berkshelf for dep resolving in ChefSpec.
Policyfile resolver for ChefSpec requires "chef-dk" gem, which has pretty big dependency tree.
So, for unit tests it's better just to use Berkshelf.
* Disable CodeClimate reporter
* Gemfile: Remove unnecessary gems
* Use Chef Delivery CLI to run spec and style checks
* Add "chef_version" to metadata

### DIFF
--- a/.delivery/project.toml
+++ b/.delivery/project.toml
@@ -1,0 +1,9 @@
+[local_phases]
+unit = "bundle exec rspec test/spec/"
+lint = 'cookstyle --display-cop-names --extra-details'
+syntax = "foodcritic ."
+provision = "echo skipping"
+deploy = "echo skipping"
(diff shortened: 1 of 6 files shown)


legal90:
@johnbellone Do we really need this `codecov` integration in CI? It doesn't look helpful at the moment, iMO.

johnbellone:
@legal90 It should be disabled as of right now.

PR #427, opened 2017-04-27
Calling /etc/init.d/consul restart sometimes ends up with a stopped …

MichaelKueller:
…consul process. The reason is that the init script sometimes starts consul before it was completely stopped. During a cookbook upgrade from version 2.1.3 to 2.2.0 this took down the consul agent on a number of our servers. 
This can be easily reproduced runnning:

 watch -n 0.5 -e '/etc/init.d/consul restart; sleep 0.1; /etc/init.d/consul status'

Adding the --retry parameter lets start-stop-daemon wait until the consul process is actually stopped.

### DIFF
--- a/templates/default/sysvinit.service.erb
+++ b/templates/default/sysvinit.service.erb
@@ -39,7 +39,7 @@ _start() {
 }
 
 _stop() {
-  start-stop-daemon --stop --quiet --pidfile $pidfile --user $user --signal "<%= @stop_signal %>"
+  start-stop-daemon --stop --quiet --pidfile $pidfile --user $user --retry="<%= @stop_signal %>"/30/KILL/5
 }
 
 _status() {


MichaelKueller:
The underlying has the potential to take down 20 percent of all consul agents when appliying a config change accross an entire environment. When hundreds of servers are involved easily dozens of them will be stopped instead of restarted. Which is why I believe this fix is quite important.

libraries/consul_config.rb
```diff
@@ -27,6 +27,9 @@ class ConsulConfig < Chef::Resource
       # @!attribute config_dir
       # @return [String]
       attribute(:config_dir, kind_of: String, default: lazy { node['consul']['service']['config_dir'] })
+      # @!attribute config_dir_mode
+      # @return [String]
+      attribute(:config_dir_mode, kind_of: String, default: '0755')
```
Ginja: This should be `0750` as there may be a  Consul Gossip encrypt key in there.

PR #430, opened 2017-05-05
Adding the option to modify the permissions on the consul conf.d dire…

MichaelKueller:
…ctory.

### DIFF
--- a/libraries/consul_config.rb
+++ b/libraries/consul_config.rb
@@ -27,6 +27,9 @@ module ConsulCookbook
       # @!attribute config_dir
       # @return [String]
       attribute(:config_dir, kind_of: String, default: lazy { node['consul']['service']['config_dir'] })
+      # @!attribute config_dir_mode
+      # @return [String]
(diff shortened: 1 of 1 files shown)


MichaelKueller:
@Ginja could you readd your comments to the pull request please. I wanted to have two seperate pull requests for the two changes I made, as they are not related to each other.

legal90:
@MichaelKueller Could you please give an example of the use case - when people might want to set custom permissions mode for this dir?

MichaelKueller:
@legal90 yes, sure. 
In our case we have a user different than the consul user, that we use to perform deployments. During the deployment this user writes the service_definition_file into /etc/consul/conf.d directory. 
With the current persmissions (0755) this is not possible, as only the consul user could do so. In this case it is very helpful if these permissions are configurable.

legal90:
Thanks, @MichaelKueller! Now it makes sense 👍 

@johnbellone, Do you have any concerns to get it merged?

PR #433, opened 2017-05-08
Improved the _start()  method in the sysv initscript. Until now a sta…

MichaelKueller:
…rt of the consul agent followed by a reload will interrupt the start process and the consul agent will be stopped

You can simply reproduce this by calling:
/etc/init.d/consul restart; /etc/init.d/consul reload; /etc/init.d/consul status

 This is likely to happen if you make a change in a wrapper cookbook that triggeres a consul restart and another change which triggeres a consul reload. Your consul agent will be stopped afterwards.

  To fix this this change will make sure the consul service is started before leaving the _start() method.

### DIFF
--- a/templates/default/sysvinit.service.erb
+++ b/templates/default/sysvinit.service.erb
@@ -32,10 +32,27 @@ export PATH="${PATH:+$PATH:}/usr/sbin:/sbin"
 _start() {
   touch $logfile
   chown $user $logfile
+  echo "Starting consul"
   start-stop-daemon --start --quiet --background \
       --pidfile $pidfile<% unless @pid_file_external %> --make-pidfile<% end %> \
       --chuid $user --chdir "<%= @directory %>" \
       --startas /bin/bash -- -c "exec $exec <%= @daemon_options %> >> $logfile 2>&1"
+
+  echo -n "Waiting for consul daemon to be listening..."
+  for i in `seq 1 30`; do
+    if ! start-stop-daemon --quiet --stop --test --pidfile $pidfile --user $user; then
+      echo " FAIL: consul process died"
+      return 2
+    fi
+    if "$exec" info >/dev/null; then
+      echo " OK"
+      return 0
+    fi
+    echo -n .
+    sleep 1
+  done
+  echo " FAIL: consul process is alive, but is not listening."
+  return 2
 }
 
 _stop() {
--- a/templates/default/sysvinit.service.erb
+++ b/templates/default/sysvinit.service.erb
@@ -112,6 +112,7 @@ _status_q() {
 _wait_for_listening() {
     echo -n "Waiting for consul daemon to be listening..."
     for i in `seq 1 30`; do
+        sleep 1
         # if ! start-stop-daemon --quiet --stop --test --pidfile $pidfile --user $user; then
         if ! _status_q; then
             echo " FAIL: consul process died"
@@ -122,7 +123,6 @@ _wait_for_listening() {
             return 0
         fi
         echo -n .
-        sleep 1
     done
     echo " FAIL: consul process is alive, but is not listening."
     return 2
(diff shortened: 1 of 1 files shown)


legal90:
@MichaelKueller Thank you!
But I see  that this fix covers only Debian platform. Could you please make it universal by moving this hook to a standalone function, like `_wait_for_listening`?

templates/default/sysvinit.service.erb
```diff
@@ -32,10 +32,27 @@ export PATH="${PATH:+$PATH:}/usr/sbin:/sbin"
...
   chown $user $logfile
+  echo "Starting consul"
   start-stop-daemon --start --quiet --background \
       --pidfile $pidfile<% unless @pid_file_external %> --make-pidfile<% end %> \
       --chuid $user --chdir "<%= @directory %>" \
       --startas /bin/bash -- -c "exec $exec <%= @daemon_options %> >> $logfile 2>&1"
+
+  echo -n "Waiting for consul daemon to be listening..."
+  for i in `seq 1 30`; do
+    if ! start-stop-daemon --quiet --stop --test --pidfile $pidfile --user $user; then
```
legal90: Could we use the existing `_status_q()` function here?
```shell
#...
if ! _status_q; then
  echo " FAIL: consul process died"
  return 2
fi
#...

**Comment by MichaelKueller** [comment]
@legal90 good point! I changed the code accordingly.

**Comment by MichaelKueller** [comment]
@legal90 I moved the hook to a _wait_for_listening function. I did not call it from the non-debian _start function though. After testing this with centos it turns out that the underlying issue does not exist there.

File: templates/default/sysvinit.service.erb [hunk 2]
```diff
@@ -106,6 +109,25 @@ _status_q() {
     _status >/dev/null 2>&1
 }
 
+_wait_for_listening() {
+  echo -n "Waiting for consul daemon to be listening..."
+  for i in `seq 1 30`; do
+    # if ! start-stop-daemon --quiet --stop --test --pidfile $pidfile --user $user; then
+    if ! _status_q; then
```
legal90: It appeared that both of these checks (`_status_q ` and the previous long version) fails if they are running right after the daemon start. I've reproduced that in Debian 7 via Test Kitchen:
```console
# service consul restart
Starting consul
Waiting for consul daemon to be listening... FAIL: consul process died
```

So, it might be reasonable to add a sleep before the first check.
For example, it could be done by moving `sleep 1` from the end to the beginning of the "for" loop:
```bash
_wait_for_listening() {
    echo -n "Waiting for consul daemon to be listening..."
    for i in `seq 1 30`; do
        sleep 1
        if ! _status_q; then
            echo " FAIL: consul process died"
            return 2
        fi
        if "$exec" info >/dev/null; then
            echo " OK"
            return 0
        fi
        echo -n .
    done
    echo " FAIL: consul process is alive, but is not listening."
    return 2
}
```

P.s. Also, the indentation size is 4 spaces in this file 😉

ZbigniewZabost-zanox: Hey

I was not able to reproduce problem you described with debian 7 (also using kitchen), but I moved sleep any way to beginning of loop.

Also, I fixed indentation.

legal90:
@MichaelKueller @ZbigniewZabost-zanox Thank you!
_________
@johnbellone Could you please disable Codecov integration for pull-requests?

legal90:
@MichaelKueller @ZbigniewZabost-zanox 
Hi guys! Seems like the method we've introduced here will not work with Consul 0.8+ having ACL enabled. In this case `consul info` could not de used for checking a service status because it always returns "403 Permission Denied", requesting the ACL token to be passed there.
Here is the test failure on Debian 7: https://travis-ci.org/johnbellone/consul-cookbook/jobs/235152247 (from PR: https://github.com/johnbellone/consul-cookbook/pull/434)

Do you have any ideas how we can avoid using `consul info` in init scripts ?

moofish32:
@legal90 -- I think we have to move the ACL into the environment of the process manager scripts. The v8 ACL touches all commands AFAIK. The variable would be `CONSUL_HTTP_TOKEN` I think. The other option is prevent the cookbook from calling restart followed by reload and add a note for the user. The issue I opened #438 is a duplicate of the problem here.  What are the key use cases requiring reload for current consumers? As 0.8 is a fairly serious change (with the raft update), perhaps this is a good opportunity to roll a major here as well and remove the reload.

PR #434, opened 2017-05-15
Update default version to 0.8.3

Ginja:
* Add disable_host_node_id config option
  * Change default value for disable_remote_exec to true to reflect new default value in Consul 0.8+
* Add checksums for Consul 0.8.2
* Remove rpc config port as the interface has been deprecated in Consul 0.8+

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -22,7 +22,6 @@ default['consul']['config']['client_addr'] = '0.0.0.0'
 default['consul']['config']['ports'] = {
   'dns'      => 8600,
   'http'     => 8500,
-  'rpc'      => 8400,
   'serf_lan' => 8301,
   'serf_wan' => 8302,
   'server'   => 8300,
@@ -32,7 +31,7 @@ default['consul']['diplomat_version'] = nil
 
 default['consul']['service']['config_dir'] = join_path config_prefix_path, 'conf.d'
 
-default['consul']['version'] = '0.7.5'
+default['consul']['version'] = '0.8.3'
 
 # Windows only
 default['consul']['service']['nssm_params'] = {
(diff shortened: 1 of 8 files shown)


legal90:
@Ginja Travis CI build failure points to something broken in the Consul runtime. That's what I've got in the Consul output while after converging `default-centos-72` Test Kitchen suite.

Seems like we need to tune some ACL settings for test environment. Could you please take a look?

Ginja:
@legal90 Done. The tests pass for me locally, and the 3 that failed on Travis didn't fail outright, they just timed out.

legal90:
@Ginja Tests on Upstart-driven systems are not passed. There is an infinite loop here after the service start:
https://github.com/johnbellone/consul-cookbook/blob/master/templates/default/upstart.service.erb#L51
It's also reproducible in Vagrant (without Docker):

```
root@ubuntu-14:~# ps aux | grep conful
<...>
root      2119  0.0  0.1  18176  1172 ?        Ss   05:26   0:00 /sbin/start consul
consul    2122  0.3  1.6  33476 16256 ?        Ssl  05:26   0:00 /opt/consul/0.8.3/consul agent -config-file=/etc/consul/consul.json -config-dir=/etc/consul/conf.d
consul    2126  0.0  0.0   4452   656 ?        Ss   05:26   0:00 /bin/sh -e -c while ! /opt/consul/0.8.3/consul info ; do sleep 1; done /bin/sh
consul    3721  0.0  0.0   4352   356 ?        S    05:30   0:00 sleep 1
<...>

root@ubuntu-14:~# /opt/consul/0.8.3/consul info
Error querying agent: Unexpected response code: 403 (Permission denied)
```

It seems like in Consul 0.8 `consul info` needs token to be passed. Otherwise, it returns non-zero exit code.

Ginja:
Thanks for fixing that, @legal90. First day back from vacation :)

Just ran `converge` for `default-ubuntu-1204` and everything looks good to me 👍 . I'll let you merge this.

legal90:
It's not completely fixed yet because of the failure on sysvinit-driven Debain distributions (Debian 7, in particular). There is one more point where we use `consul info` call, which causes a 403 failure on Consul 0.8 with ACL enabled:
https://travis-ci.org/johnbellone/consul-cookbook/jobs/235152247

I'll try to figure out how we can fix it.

PR #441, opened 2017-05-24
Fix /opt/consul permissions

moofish32:
The recursive call on the directory in service provider overrides the configuration previously setup for the /opt/consul directory

### DIFF
--- a/libraries/consul_service.rb
+++ b/libraries/consul_service.rb
@@ -76,7 +76,7 @@ module ConsulCookbook
       def action_enable
         notifying_block do
           directory new_resource.data_dir do
-            recursive true
+            recursive false
             owner new_resource.user
             group new_resource.group
             mode '0750'


legal90:
Hi @moofish32,
You probably missed something, because `data_dir` param of `consul_service` resource defaults to `/var/lib/consul`:
https://github.com/johnbellone/consul-cookbook/blob/d1a4af3bb2d7b26c57df82f7671e019f9df4601e/libraries/helpers.rb#L44

So, it doesn't not affect consul installation path `/opt/consul`, which is managed in `consul_installation` resource:
https://github.com/johnbellone/consul-cookbook/blob/master/libraries/consul_installation_binary.rb#L45

legal90:
Also, `recursive` param of `directory` resource only means that the whole directory structure will be created.  It doesn't affect permissions mode. E.q. it is _not_ an analogue of (`chmod -R`)
https://docs.chef.io/resource_directory.html#recursive-directories

PR #442, opened 2017-05-25
modify consul_acl resource to support SSL options

shayangz:
Currently the `consul_acl` resource breaks with an SSL only Consul cluster. This PR adds ability to pass optional SSL attributes to enable using the resource in such situations.

### DIFF
--- a/libraries/consul_acl.rb
+++ b/libraries/consul_acl.rb
@@ -39,6 +39,26 @@ module ConsulCookbook
       # @return [String]
       attribute(:rules, kind_of: String, default: '')
 
+      # @!attribute client_cert
+      # @return [String]
+      attribute(:client_cert, kind_of: String, default: nil)
+
+      # @!attribute client_key
+      # @return [String]
+      attribute(:client_key, kind_of: String, default: nil)
+
+      # @!attribute ca_file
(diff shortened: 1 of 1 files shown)


legal90:
Hi @shayangz,
Thank you for the contribution! Adding SSL options to `consul_acl` is a good idea. But since `ssl` hash is passing directly to Faraday HTTP lib, may be it would be better an easier just to add a single attribute `ssl` to allow users to pass the desired hash?

```ruby
      # @!attribute ssl
      # @return [Hash]

      attribute(:ssl, kind_of: Hash, default: {})
```

Then it should be possible to pass a hash with custom SSL settings:
```ruby
consul_acl 'token' do
  # ... other params
  ssl({
    client_cert: '...',
    client_key: '...',
    ca_file: '...',
    ca_path: '...',
    cert_store: '...'
  })
end
```
as well as this one (to disable peer verification):
```ruby
consul_acl 'token' do
  # ... other params
  ssl({ verify: false })
end
```

shayangz:
Consolidating the options into a single Hash sounds great. Updated the PR.

legal90:
@shayangz Thank you!

PR #445, opened 2017-05-30
Use "service_user" and "service_group" attributes as defaults for resource params 

legal90:
It looks like a mistake existing for a long time already.
Regardless of their names, attributes `['consul']['service_user']` and `['consul']['service_group']` don't actually affect the user running Consul as well as permissions for `['consul']['config']['data_dir']` directory.

This PR fixes that. Now `user` and `group` params of `consul_service` resource have default values as `['consul']['service_user']` and `['consul']['service_group']` appropriately.

### DIFF
--- a/libraries/consul_service.rb
+++ b/libraries/consul_service.rb
@@ -24,11 +24,11 @@ module ConsulCookbook
       # @!attribute user
       # The service user the Consul process runs as.
       # @return [String]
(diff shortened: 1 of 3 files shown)


legal90:
I've added the same for `consul_watch` resource as well, so now it behaves similar to `consul_definitions` (in terms of owner&group of the file generated)

PR #447, opened 2017-05-31
consul_config: Remove default values

legal90:
We don't need to set these defaults values because:
1) These default values are rendered to the `consul.json` config file only if an appropriate argument was called in the `consul_config` resource source code. For example, currently we have `verify_incoming` and `verify_outgoing` always rendered in the config file because they are called here: https://github.com/johnbellone/consul-cookbook/blob/master/libraries/consul_config.rb#L208

2) Consul can handle default values by itself. 
We just don't need to render them in the config file unless we really need to set them explicitly. It makes a big sense for consul config options which have different default values in different Consul agent versions. For example, `acl_enforce_version_8` defaults to _false_ in versions of Consul prior to 0.8, and defaults to _true_ in Consul 0.8 and later.

This PR doesn't break the backward compatibility. No explicitly defined variables will be removed.


legal90:
I've fixed the spec as well. 
As we see, now the config file will contain only those options, which have been explicitly set by resource param and/or node's attributes `['consul']['config'][<opt_name>]`

legal90:
@johnbellone Could you please disable codecov integration for Pull-Requests in this repo? 
We can still use it for a badge in README.md, but let's keep PRs green.

PR #449, opened 2017-05-31
Fix consul application path on Windows

slavirok:
There was a hard coded Unix path which was being used on Windows Installation.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -42,3 +42,9 @@ default['consul']['service']['nssm_params'] = {
   'AppRotateOnline'  => 1,
   'AppRotateBytes'   => 20_000_000,
 }
+
+if node['os'].eql?('windows')
+  node.default['consul']['service']['program'] = join_path program_files, 'consul', node['consul']['version'], 'consul.exe'
+else
+  node.default['consul']['service']['program'] = '/usr/local/bin/consul'
+end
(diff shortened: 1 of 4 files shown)


legal90:
Hi @slavirok ! Could you please evaluate - what was the issue you have faced with? Do you install Consul from a package (msi?), or just as a binary?
Sharing the recipe code would be great.

legal90:
The `package` provider for `consul_installation` resource has UNIX-specific path just because it doesn't support Windows yet.
The expected behavior for Windows is using the default provider (`binary`), as it is done in `consul::default` recipe:
https://github.com/johnbellone/consul-cookbook/blob/eebc1299cefb6d766218904ea372566cc84bf373/recipes/default.rb#L23-L40

slavirok:
Hi @legal90 
this is my recipe:

```
include_recipe 'consul::default'

consul_service 'consul' do
  notifies :restart, 'service[consul]'
end

service 'consul' do
  action :nothing
end
```

legal90:
@slavirok This recipe looks wrong. You should not call `consul_service` resource since it is already called in `'consul::default'` recipe.

What is the expected behavior? Do you want to **restart** (instead of reload) the `consul` service every time the `/etc/consul/consul.json` is changed?

legal90:
Closing this PR since it's not relevant to the issue described here.

PR #456, opened 2017-07-17
Fix foodcritic violations

Annih:
- Use proper license name
- Use converge_by instead of updated_by_last_action
- Use platform introspection DSL when possible
- Ignore FC028 when DSL is not usable due to poise inversion

*Cc.* @aboten

### DIFF
--- a/libraries/consul_installation_binary.rb
+++ b/libraries/consul_installation_binary.rb
@@ -32,7 +32,7 @@ module ConsulCookbook
       # @return [Hash]
       # @api private
       def self.default_inversion_options(node, resource)
-        extract_path = node.platform_family?('windows') ? node.config_prefix_path : '/opt/consul'
+        extract_path = windows? ? node.config_prefix_path : '/opt/consul'
         super.merge(extract_to: extract_path,
                     version: resource.version,
                     archive_url: 'https://releases.hashicorp.com/consul/%{version}/%{basename}',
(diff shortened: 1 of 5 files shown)


libraries/consul_installation_binary.rb
```diff
@@ -32,7 +32,7 @@ def self.provides_auto?(_node, _resource)
       # @return [Hash]
       # @api private
       def self.default_inversion_options(node, resource)
-        extract_path = node.platform_family?('windows') ? node.config_prefix_path : '/opt/consul'
+        extract_path = windows? ? node.config_prefix_path : '/opt/consul'
```
legal90: Seems like this hunk breaks the converge:
https://travis-ci.org/johnbellone/consul-cookbook/builds/254804735
```
undefined method `windows?' for ConsulCookbook::Provider::ConsulInstallationBinary:Class
```

Could you please try to replace the condition with `platform?('windows')` , as we do in other resources here?

recipes/default.rb
```diff
@@ -9,7 +9,7 @@
 poise_service_user node['consul']['service_user'] do
   group node['consul']['service_group']
   shell node['consul']['service_shell'] unless node['consul']['service_shell'].nil?
-  not_if { node.platform_family?('windows') }
```
legal90: `platform_family?` method is not available in Chef <= 12.6.0 (https://github.com/johnbellone/consul-cookbook/pull/417).
Let's just keep it as is and add `FC028` to the ignore list in `.foodcritic` file.

Annih:
Foodcritic violations have been fixed by another review :)

Issue #451, opened 2017-07-05
Consul command not available on Windows


emachnic:
Just a quick fix but currently, the Windows installation does not add the Consul directory to the path and trying to run `consul` commands from Powershell do not work. I was able to get around this by using `windows_path` from the Windows cookbook like so:

```ruby
windows_path "C:/Program Files/consul/#{node['consul']['version']}" do
  action :add
end
```

It might be cool to go ahead and do this in the cookbook already so that the consumers of the cookbook don't have to.

Cheers,
Evan

gdavison:
A shortcut might be a better option, in the same way that the *nix versions use a symlink. I haven't tested that this works, though

gdavison:
@emachnic, a Windows symlink will do it. I have code on https://github.com/paybyphone/consul-cookbook/tree/windows_executable_symlink that creates it.

It's waiting on #452 to fix Windows issues with cookbook version 3.0

gdavison:
@emachnic This is available on the `master` branch, but not yet in a numbered release

damacus:
Closing due to inactivity. 

If this is still an issue please reopen or open another issue. Alternatively drop by the #sous-chefs channel on the [Chef Community Slack](http://community-slack.chef.io/) and we'll be happy to help! 

Thanks,
Sous-Chefs

PR #460, opened 2017-07-21
Windows executable symlink

gdavison:
Creates a symlink to the current Consul version at `C:\Program Files\consul\consul.exe`

Addresses #451


legal90:
@gdavison,
Thank you for this Pull-Request! 👍 
As I see, `windows_path` is a pretty new resource, available only in Chef Client v13.4 and later.
Do you know - is there any other way to do the same on the earlier version of Chef Client?

gdavison:
@legal90 The `windows_path` function was previously provided by the `windows` cookbook, then added to core Chef in v13.4. https://github.com/chef-cookbooks/windows#deprecated-resources-note

I'm going to update the cookbook code and explicitly include the `windows` cookbook as a dependency.

libraries/consul_installation_binary.rb
```diff
@@ -59,6 +59,15 @@ def action_create
             to ::File.join(options[:extract_to], new_resource.version, 'consul')
             not_if { windows? }
           end
+
+          link "#{node.config_prefix_path}\\consul.exe" do
+            to ::File.join(options[:extract_to], new_resource.version, 'consul.exe')
+            only_if { windows? }
+          end
+          windows_path node.config_prefix_path do
```
legal90: Please, try if we can just use `env` instead:
```
  env 'path' do
    action :modify
    delim ::File::PATH_SEPARATOR
    value node.config_prefix_path
    only_if { windows? }
  end
```
It should work as expected, on Chef >= 12, without having to add a `windows` dependency

gdavison: I can certainly take a look.

But the `windows` cookbook isn't a new dependency: the `consul` cookbook depends on `build-essential`, which depends on `seven-zip`, which has a dependency on `windows`. So this is more making it an explicit dependency. And with `windows_path` becoming part of core Chef, it seems more clear going forward.

Let me know what you think.

legal90: > the `consul` cookbook depends on `build-essential`, which depends on `seven-zip`, which has a dependency on `windows`

Ah, sorry, I forgot about that. Then I'm OK with setting that dependency explicitly. 👍

gdavison: No worries, I had the "wait, why did that work" moment myself :)

PR #466, opened 2017-08-24
add flags attribute

sigmatechnica:
Consul needs runtime flags for some stuff (like running the UI).  This PR adds a flags attribute that allows that, and by default sets to to nothing.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -28,7 +28,7 @@ default['consul']['config']['ports'] = {
 default['consul']['diplomat_version'] = nil
 
 default['consul']['service']['config_dir'] = join_path config_prefix_path, 'conf.d'
-default['consul']['service']['flags'] = ""
+default['consul']['service']['flags'] = ''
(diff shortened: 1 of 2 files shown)


legal90:
Hi @sigmatechnica,
In this cookbook, we try to keep the only one way for passing configuration parameters to Consul agent. It's the main configuration file, managed by `consul_config` resource.

In particular, UI could be enabled by setting an attribute:
```ruby
node.default['consul']['config']['ui'] = true
```

So, I would prefer to do not have flag support to avoid confusions when the same parameter could be defined in multiple places.

@johnbellone @Ginja, what do you think about it?

sigmatechnica:
Aah didn't find node.default['consul']['config']['ui'] documented anywhere so invented my own way.  agree, does make this a bit pointless

sigmatechnica:
closing as it's unneeded (and fails CI checks for centos 7 anyway)

Issue #474, opened 2017-10-23
Custom resources do not implement to_json correctly


Wing924:
to_json interface should be `def to_json(options = nil)` but was `def to_json`
It causes:
```
 #<ArgumentError: wrong number of arguments (given 1, expected 0)>
    /var/chef/cache/cookbooks/consul/libraries/consul_config.rb:133:in `to_json'
```

legal90:
AFAIK, it's not supposed that this method will be called with arguments. When does this issue happen?

Wing924:
@legal90 `JSON.generate` or `.to_json` will call that method with args.

legal90:
`to_json` is a private method here and it's called without any arguments. All tests are passing fine.
Please, provide a scenario how can we reproduce this issue.

Wing924:
@legal90 
Create a Chef Handler like this:
```
class Chef; class Handler
	def report
                # The data method is used to return the Hash representation of the run_status object

		# these lines will raise error
		data.to_json 
		JSON.generate data
	end
end; end
```
ref. https://docs.chef.io/handlers.html

legal90:
OK, now I got it.
So, the issue is caused by the method name collision. The proper way to fix it is just renaming our `to_json` methods to something unique. For example, to `params_to_json`.

I will send a PR.

legal90:
@Wing924 Please, verify your handler with this fix applied: https://github.com/johnbellone/consul-cookbook/pull/476

PR #475, opened 2017-10-23
Fix: Custom resources do not implement to_json correctly

Wing924:
Fix: #474


Wing924:
@johnbellone The test was failed but not from my codes. Please check it.

PR #478, opened 2017-11-02
Add some missing config options

jescholl:
Hashicorp has added some new config options.  This adds the missing options and adds some additional indicators to the `tls?` method.

Full documentation available at https://www.consul.io/docs/agent/options.html#configuration-files.


Lasering:
Can I help moving this PR forward? I would like to use the `ca_path` config.

legal90:
@jescholl Thank you!

Issue #461, opened 2017-07-24
Race condition when acquiring cluster leadership for tests


gdavison:
I've noticed a number of times when running the integration tests that I get the following error with the ACL tests:

```
Diplomat::UnknownStatus
-----------------------
status 500: No cluster leader
```

I assume that it's a race condition between the server initializing itself and the ACL test being run. It's currently affecting PR #460 

@legal90, would it be worthwhile adding a `retry` to the Diplomat requests? I can put together a PR for this

legal90:
Hi @gdavison,
> would it be worthwhile adding a retry to the Diplomat requests? I can put together a PR for this

Yeah, that would be great! 👍

PR #482, opened 2018-01-05
Add consul API retries

gdavison:
When starting up a new machine and creating an ACL, the Consul server is not always ready to serve requests by the time the ACL is ready to be created. This causes intermittent failures in integration tests, and likely in real-world use.

Two retries with an interval of 0.5 seconds seems to work.

Also addressed some `cookstyle` issues


libraries/consul_acl.rb
```diff
@@ -90,10 +90,35 @@ def configure_diplomat
...
+        begin
+          return yield attempts
+        rescue Diplomat::UnknownStatus
+          attempts += 1
+
+          # If we've maxed out our attempts, raise the exception to the calling code
+          raise if attempts > opts[:attempts]
+
+          # Sleep before the next retry if the option was given
+          sleep 0.5
```
legal90: Should it be `sleep opts[:sleep]` instead?

libraries/consul_acl.rb
```diff
@@ -90,10 +90,35 @@ def configure_diplomat
...
+          return false if old_acl.nil? || old_acl.empty?
+          old_acl.first.select! { |k, _v| %w(ID Type Name Rules).include?(k) }
+          old_acl.first == new_resource.to_acl
+        end
+      end
+
+      def retry_block(opts = {}, &_block)
+        opts = {
+          sleep: 0, # Seconds to sleep between attempts
+        }.merge(opts)
```
legal90: I think, we should also define a define for `:attempts` option here. Otherwise, if it's not passed to the method, `opts[:attempts]` will be `nil` and line 115 will fail:
```
ArgumentError: comparison of Fixnum with nil failed
```

libraries/consul_acl.rb
```diff
@@ -90,10 +90,35 @@ def configure_diplomat
...
+
+        opts[:catch] = [opts[:catch]].flatten
+        attempts = 1
+
+        begin
+          return yield attempts
+        rescue Diplomat::UnknownStatus
```
legal90: Should it be `rescue *opts[:catch]` instead?

libraries/consul_acl.rb
gdavison: Yes :)

libraries/consul_acl.rb
gdavison: No, this will be the only error we want to retry on

libraries/consul_acl.rb
gdavison: Yes

gdavison:
@legal90 I've removed some bits I should have removed when I copied the code. It was originally much more general, but I've made it specific here.

I haven't tried it for real, because we don't use ACLs yet. My main reason for adding this was to remove the intermittent test failures. I've added an additional attempt, so it will give 1.5 seconds for the consul service to start. It can definitely be tuned if needed.

libraries/consul_acl.rb
```diff
@@ -90,10 +90,35 @@ def configure_diplomat
...
+        end
+      end
+
+      def retry_block(opts = {}, &_block)
+        opts = {
+          attempts: 3, # Number of attempts
+          sleep:    0, # Seconds to sleep between attempts
+        }.merge(opts)
+
+        attempts = 1
```
legal90: I would suggest renaming `attempts` to something like `count`, or `opt[:attempts]` to `opt[:max_attempts]` in order to avoid the confusion between these two variables.

libraries/consul_acl.rb
legal90: Yeah, now it's fine

gdavison:
Hi @johnbellone, any feedback on the PR?

gdavison:
Fixes #461

gdavison:
@johnbellone, yes, it would (still) fail in this case if the local Consul agent were not yet available. Before this change, there would be a race condition which caused intermittent failures, at least is the tests. Now, the behaviour is to try three times before failing.

PR #488, opened 2018-02-11
Fix git install integration tests

gdavison:
Updates to allow git-based installations.

* Updates the installed `go` version to 1.9, which is the current required minimum
* Adds the `apt` cookbook to update packages for Debian and Ubuntu
* Removes some end-of-life Linux distros.


metadata.rb
```diff
@@ -16,6 +16,7 @@
 supports 'arch'
 supports 'windows'
 
+depends 'apt'
```
legal90: I'm not sure that `apt` should be added as a dependency for this cookbook. If it's needed for tests, then let's add it to `test/fixtures/policies/default.rb`:
```
cookbook 'apt'
```

test/fixtures/cookbooks/consul_spec/recipes/default.rb
```diff
@@ -3,6 +3,8 @@
 # this allows the consul_definition recipe to run correctly
 node.default['consul']['config']['enable_script_checks'] = true
 
+include_recipe 'apt::default' if node.platform_family?('debian')
```
legal90: Are tests failing without this line?

gdavison:
Yeah, kitchen 1.20.0 seems to have an issue with a missing dependency. I got the same error at home until I updated.

gdavison:
I think the Codecov error is spurious, @johnbellone @legal90

Issue #498, opened 2018-06-02
consul 3.1.0 issue when testing with hashicorp-vault 3.0.0 git code


knightorc:
I'm investigating this issue.

Testing pre release hashicorp-vault and I'm seeing it mucking with consul, 3.0.0 and 3.1.0

```    * execute[Reload consul] action run[2018-06-02T17:44:12+00:00] INFO: Processing execute[Reload consul] action run (C:/chef/cache/cookbooks/consu
libraries/consul_service_windows.rb line 49)

      [execute] 'consul.exe' is not recognized as an internal or external command,
                operable program or batch file.

      ================================================================================
      Error executing action `run` on resource 'execute[Reload consul]'
      ================================================================================

      Mixlib::ShellOut::ShellCommandFailed
      ------------------------------------
      Expected process to exit with [0], but received '1'
      ---- Begin output of consul.exe reload ----
      STDOUT:
      STDERR: 'consul.exe' is not recognized as an internal or external command,
      operable program or batch file.
      ---- End output of consul.exe reload ----
      Ran consul.exe reload returned 1

      Cookbook Trace:
      ---------------
      C:/chef/cache/cookbooks/poise/files/halite_gem/poise/helpers/notifying_block.rb:69:in `notifying_block'
      C:/chef/cache/cookbooks/consul/libraries/consul_service_windows.rb:48:in `action_reload'

      Resource Declaration:
      ---------------------
      # In C:/chef/cache/cookbooks/consul/libraries/consul_service_windows.rb

       49:           execute 'Reload consul' do
       50:             command 'consul.exe reload' + (new_resource.acl_token ? " -token=#{new_resource.acl_token}" : '')
       51:             cwd ::File.dirname(new_resource.program)
       52:             action :run
       53:           end
       54:         end

      Compiled Resource:
      ------------------
      # Declared in C:/chef/cache/cookbooks/consul/libraries/consul_service_windows.rb:49:in `block in action_reload'

      execute("Reload consul") do
        action [:run]
        default_guard_interpreter :execute
        command "consul.exe reload"
        backup 5
        cwd "C:\\Program Files\\vault\\0.8.4"
        returns 0
        user nil
        declared_type :execute
        cookbook_name "consul"
        domain nil
      end

      System Info:
      ------------
      chef_version=13.9.1
      platform=windows
      platform_version=6.1.7601
      ruby=ruby 2.4.3p205 (2017-12-14 revision 61247) [x64-mingw32]
[...]

PR #499, opened 2018-06-02
Fix name collisions in helpers.rb

knightorc:
When a run_list includes consul and hashicorp-vault their helpers inject into the node colliding def blocks.

Depending on the order of inclusion, `consul` or `vault` is used, when it should be the other.

Solves: https://github.com/johnbellone/consul-cookbook/issues/498


knightorc:
Clossing - Reconsidering my fix.

Issue #484, opened 2018-01-15
tls_cipher_suites should be a comma separated list


Lasering:
The recipe should convert the `tls_cipher_suites` array to a CSV which is what the consul agent is expecting. Otherwise we get the following error:
```
* 'tls_cipher_suites' expected type 'string', got unconvertible type '[]interface {}'
```

mfischer90:
This still seems to be an issue, makes using this variable impossible

ragarcia26:
bump, seeing this too

mfischer90:
#501 created to fix

damacus:
Closing via #501

PR #501, opened 2018-06-18
Updating bug in configuration for consul

mfischer90:
tls_cipher_suites needs to be a comma separated string, not an array

see #484


gdavison:
@mfischer90 we now have `consul_config_v0.rb` and `consul_config_v1.rb` for v0.x and v1.x configurations. Can you make your change to both files please?

PR #496, opened 2018-05-11
Consul v1.0.x support

gdavison:
Adds conversion from pre-v1.0 fields to new v1.0 equivalents, as documented in https://github.com/hashicorp/consul/blob/master/CHANGELOG.md#100-october-16-2017, section "Deprecated Options Have Been Removed" (actually many introduced in v0.9.3).

PR #502, opened 2018-06-29
Support for discovery_max_stale

pierresouchay:
Enable support of config option available since version 1.0.7

https://www.consul.io/docs/agent/options.html#discovery_max_stale


gdavison:
Hi @pierresouchay, I recently merged #496 which separated the Consul 0.x and 1.x configuration resources, since they're diverging

pierresouchay:
Hello @gdavison ,

I now added some unit tests and allow to have the same attribute to be used independently of any Consul Version (in version where is is not supported, the attribute won't be present in config file).

Should be fine now

Kind Regards

pierresouchay:
@gdavison is the patch Ok for you?

Kind Regards

pierresouchay:
@gdavison is the patch Ok for you?

I had to fight a bit with foodcritic

pierresouchay:
@gdavison Do you have plans to include this PR ?

Regards

pierresouchay:
@johnbellone  @gdavison Would you consider this patch? Do you have some changes you would like to me to do?

More generally, we recently had an outage regarding a deprecated feature in Consul (basically consul_definition with a check using `script` instead of `args`. Since this version was removed from Consul 1.1, while upgrading to Consul 1.1, Consul agent did fail to restart, leading to an outage. I think it might be a good idea to manage such changes in the cookbook. I did it in this review (by simply ignoring the parameter), but I think it would be nice breaking Convergence for the example when such parameter is shown. We also could start logging for deprecated options (`args` was deprecated since 1.0.x). Do you have any opinion regarding this?

Kind Regards

pierresouchay:
@johnbellone @gdavison Anyone for a review ?

gdavison:
Yes, sorry. I’ll take a look today

pierresouchay:
@gdavison Thank you!

PR #508, opened 2018-10-17
WIP: Enable CircleCI

Xorima:
This PR will enable Circle CI Integration testing


damacus:
psst, you need a Dangerfile

Xorima:
Can I get some advice on the failing tests due to a bundle install not being run, how is the best way to approach this issue? (I haven't used gems before)

gdavison:
You might have some trouble with this one, it’s a Poise-based cookbook, so it probably needs lots of those gems

Xorima:
If we had a bundle install before the `chef exec delivery` do you think this would work?

damacus:
I think we actually just need to look at removing poise once you've finished with most to the other problems

Xorima:
If you want to merge this @damacus feel free to, for now I will focus on the simpler circleci cookbooks and we can revisit this one at a later date

Issue #545, opened 2020-01-27
undefined method `join_path' for PoiseArchive::Resources::PoiseArchive::Resource


jslusher:
I started getting the following error when running a chef-client on nodes using the consul cookbook:

```
  * consul_installation[1.6.2] action create

    ================================================================================
    Error executing action `create` on resource 'consul_installation[1.6.2]'
    ================================================================================

    NoMethodError
    -------------
    undefined method `join_path' for PoiseArchive::Resources::PoiseArchive::Resource

    Cookbook Trace:
    ---------------
    /var/chef/cache/cookbooks/consul/libraries/consul_installation_binary.rb:52:in `block (2 levels) in action_create'
    /var/chef/cache/cookbooks/consul/libraries/consul_installation_binary.rb:51:in `block in action_create'
    /var/chef/cache/cookbooks/poise/files/halite_gem/poise/helpers/subcontext_block.rb:54:in `instance_eval'
    /var/chef/cache/cookbooks/poise/files/halite_gem/poise/helpers/subcontext_block.rb:54:in `subcontext_block'
    /var/chef/cache/cookbooks/poise/files/halite_gem/poise/helpers/notifying_block.rb:67:in `notifying_block'
    /var/chef/cache/cookbooks/consul/libraries/consul_installation_binary.rb:44:in `action_create'
```

```
...
  - ohai (5.3.0)
  - poise-python (1.7.0)
  - poise (2.8.2)
  - poise-languages (2.1.2)
  - poise-archive (1.5.0)
 ...
  - yum-epel (3.3.0)
...
  - consul (3.3.0)
...
  - golang (1.7.0)
  - poise-service (1.5.2)
  - hashicorp-vault (3.0.2)
```
`version 14.9.13` on centos 7 in AWS

It seems to be a poise dependency. It coincides with an update of the `vault` cookbook that removes the `default` recipe and I noticed the poise deps were removed there.

ermusthofa:
is it by any chance you solved this issue ? I also have this problem.

jslusher:
I'm working around it at the moment by locking the version of this (consul) cookbook to `3.1.0`, but I suspect there needs to be a change in one of the `poise-*` dependencies before it's fixed.

ermusthofa:
Hi @jslusher, I tried to pin the consul cookbook on version 3.1.0, but it doesn't seems to temporary resolve the issue. Would you be more specific ?

I see the join_path method in consul libraries, is it possible to do the modification there ?

[...]

PR #546, opened 2020-02-04
fix undefined method 'join_path' for PoiseArchive::Resources::PoiseAr…

scalp42:
Fix for #545 (`extend self` keeps the instance methods public, but `module_function` makes them private.)

cc @tas50

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -9,7 +9,7 @@ module ConsulCookbook
   module Helpers
     include Chef::Mixin::ShellOut
 
-    module_function
+    extend self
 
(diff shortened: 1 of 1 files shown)


scalp42:
See also https://github.com/rubocop-hq/ruby-style-guide/issues/556, I'm assuming cookstyle inherits from rubocop.

tas50:
This actually came up internally. We definitely need to disable this one. Rubocop treats them the same and they are not.

tas50:
Here's the Cookstyle PR to fix this: https://github.com/chef/cookstyle/pull/517

